PR #32, opened 2019-06-09
Polish attention related modules

gpengzhi:
1. polish and simplify the attention related modules.
2. move `attention_mechanism` related utils to `attention_mechanism_utils`.
3. add `sparsemax` in `attention_mechanism_utils` which can be used as the `probability_fn` in `LuongAttention` and `BahdanauAttention`.
4. add `attention_mechanism_utils_test`
5. For each batch of the input sequence, we only need to preprocess `memory` once (Add line 633 in `rnn_decoder.py`).

### DIFF
--- a/texar/core/attention_mechanism.py
+++ b/texar/core/attention_mechanism.py
@@ -30,8 +30,8 @@ import torch
 import torch.nn.functional as F
 from torch import nn
 
-from texar.core.attention_mechanism_utils import maybe_mask_score, \
-    prepare_memory, safe_cumprod
+from texar.core.attention_mechanism_utils import (
+    maybe_mask_score, prepare_memory, safe_cumprod)
 from texar.module_base import ModuleBase
 from texar.utils.types import MaybeTuple
 
@@ -50,31 +50,26 @@ State = TypeVar('State')
 
 
 class AttentionMechanism(ModuleBase):
-
-    @property
-    def memory_layer(self):
-        raise NotImplementedError
-
-    @property
-    def query_layer(self):
-        raise NotImplementedError
-
-    @property
-    def values(self):
-        raise NotImplementedError
-
-    @property
-    def encoder_output_size(self):
-        raise NotImplementedError
-
-
-class _BaseAttentionMechanism(AttentionMechanism):
     r"""A base AttentionMechanism class providing common functionality.
 
     Common functionality includes:
-        1. Storing the query and memory layers.
-        2. Preparing the score mask value.
+
+    1. Storing the query and memory layers.
+    2. Preparing the score mask value.
+
+    Args:
+        encoder_output_size: The output size of the encoder cell.
+        memory_layer: Instance of `torch.nn.Linear`. The layer's depth must
+            match the depth of ``query_layer``.
+        query_layer (optional): Instance of `torch.nn.Linear`. The layer's
+            depth must  match the depth of ``memory_layer``.  If
+            ``query_layer`` is not provided, the shape of ``query`` must
+            match that of ``memory_layer``.
+        score_mask_value (optional): The mask value for score before
+            passing into `probability_fn`. The default is -inf. Only used
+            if `memory_sequence_length` is not None.
     """
+
     # Cached variables that are initialized by transforming the `memory` at
     # the first forward pass of each batch. `clear_cache` should be called when
     # the batch is finished to prevent holding references to variables in the
@@ -87,20 +82,6 @@ class _BaseAttentionMechanism(AttentionMechanism):
                  memory_layer: nn.Module,
                  query_layer: Optional[nn.Module] = None,
                  score_mask_value: Optional[torch.Tensor] = None):
-        r"""Construct base AttentionMechanism class.
-
-        Args:
-            encoder_output_size: The output size of the encoder cell.
-            memory_layer: Instance of `torch.nn.Linear`. The layer's depth must
-                match the depth of ``query_layer``.
-            query_layer (optional): Instance of `torch.nn.Linear`. The layer's
-                depth must  match the depth of ``memory_layer``.  If
-                ``query_layer`` is not provided, the shape of ``query`` must
-                match that of ``memory_layer``.
-            score_mask_value (optional): The mask value for score before
-                passing into `probability_fn`. The default is -inf. Only used
-                if `memory_sequence_length` is not None.
-        """
         super().__init__()
 
         if (query_layer is not None and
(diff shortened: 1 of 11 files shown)


texar/core/cell_wrappers.py
```diff
@@ -715,7 +712,7 @@ def forward(self,  # type: ignore
                     previous_alignment_history is not None):
```
huzecong: Should this be `previous_alignment_history[i] is not None`?

texar/core/cell_wrappers.py
```diff
@@ -596,19 +597,13 @@ def zero_state(self,
                 :class:`~texar.modules.BeamSearchDecoder`.
 
         Args:
-            batch_size: `0D` integer tensor: the batch size.
+            batch_size: `0D` integer: the batch size.
 
         Returns:
             An :class:`~texar.core.AttentionWrapperState` tuple containing
             zeroed out tensors and, possibly, empty `TensorArray` objects.
```
huzecong: We don't have `TensorArray`s, those should be just plain Python lists.

texar/core/attention_mechanism.py
```diff
@@ -330,24 +275,21 @@ def __init__(self,
...
             probability_fn = lambda x: F.softmax(x, dim=-1)
         self.wrapped_probability_fn = lambda score, _: probability_fn(score)
+
         super(LuongAttention, self).__init__(
-            query_layer=None,
             memory_layer=nn.Linear(encoder_output_size, num_units, False),
+            query_layer=None,
             score_mask_value=score_mask_value)
-        self._num_units = num_units
         self._encoder_output_size = encoder_output_size
```
huzecong: Since `encoder_output_size` is set in all subclasses, it might be better to make it an argument of the base class constructor, so we could remove `Optional` from the type.

texar/core/attention_mechanism_utils.py
```diff
@@ -0,0 +1,225 @@
...
+    Returns:
+        A (possibly masked), new ``memory``.
+
+    Raises:
+        ValueError: if ``memory`` and ``memory_sequence_length`` do not have
+        the same ``batch_size``.
+    """
+    if (memory_sequence_length is not None and
+            not isinstance(memory_sequence_length, torch.Tensor)):
+        memory_sequence_length = torch.tensor(memory_sequence_length,
```
huzecong: This tensor should have the same `device` as `memory`.

texar/core/attention_mechanism_utils.py
```diff
@@ -0,0 +1,225 @@
...
+        Cumulative product of x.
+    """
+    if not isinstance(x, torch.Tensor):
+        x = torch.tensor(x)
+
+    type_map = {torch.float16: np.float16,
+                torch.float32: np.float32,
+                torch.float64: np.float64}
+
+    tiny = np.finfo(type_map[x.dtype]).tiny
```
huzecong: PyTorch has `finfo` as well.

texar/core/attention_mechanism_utils.py
```diff
@@ -0,0 +1,225 @@
...
+                                  *args, **kwargs))
+
+
+def _make_ix_like(input: torch.Tensor,
+                  dim: int = -1) -> torch.Tensor:
+    d = input.size(dim)
+    rho = torch.arange(1, d + 1, device=input.device, dtype=input.dtype)
+    view = [1] * input.dim()
+    view[0] = -1
+    return rho.view(tuple(view)).transpose(0, dim)
```
huzecong: Instead of casting `view` to a tuple, it could be created as a tuple:
```python
view = (-1,) + (1,) * (input.dim() - 1)
```

texar/core/attention_mechanism_utils.py
```diff
@@ -0,0 +1,225 @@
...
+    extra_ones = [1] * (rank - 2)
+    m_batch_size = memory.shape[0]
+
+    if seq_len_mask is not None:
+        if seq_len_batch_size != m_batch_size:
+            raise ValueError("memory_sequence_length and memory tensor "
+                             "batch sizes do not match.")
+        seq_len_mask = torch.reshape(
+            seq_len_mask,
+            tuple(list(seq_len_mask.shape) + extra_ones))
```
huzecong: This cast could also be replaced by tuple construction:
```python
seq_len_mask.size() + (1,) * (rank - 2)
```
And maybe `torch.reshape` could be replaced by `seq_len_mask.view`?

texar/modules/decoders/rnn_decoders.py
```diff
@@ -628,9 +626,15 @@ def forward(self,  # type: ignore
...
             max_decoding_length, impute_finished)
 
+        # Release memory and memory_sequence_length in AttentionRNNDecoder
         self.memory = None
         self.memory_sequence_length = None
 
+        # Release memory and memory_sequence_length in AttentionMechanism
+        for attention_mechanism in self._cell._attention_mechanisms:
+            attention_mechanism._values = None
+            attention_mechanism._keys = None
```
huzecong: Nice catch here! For design purposed, I would suggest making these two lines a method of `AttentionMechanism` (called `_clear_cache` or something). This way we'll remember to modify the release code if we add or remove new variables in `AttentionMechanism`. Also, we can document the behavior (and the dependency on `RNNDecoder` to clear the cache) in the docstring of that method.

texar/core/attention_mechanism.py
```diff
@@ -846,46 +757,43 @@ def forward(self,  # type: ignore
...
 
         Returns:
-            Tensor of dtype matching `self.values` and shape
-            `[batch_size, alignments_size]` (`alignments_size` is memory's
-            `max_time`).
+            Tensor of dtype matching ``memory`` and shape
+            ``[batch_size, alignments_size]`` (``alignments_size`` is memory's
+            ``max_time``).
         """
         processed_query = self.query_layer(query) if self.query_layer else query
```
huzecong: Since this part is almost the same for the 4 types of attention mechanisms, consider moving it to the base class `forward` method so we can share code.

texar/core/attention_mechanism.py
```diff
@@ -110,9 +112,36 @@ def __init__(self,
...
+        Args:
+            query: tensor, shaped ``[batch_size, query_depth]``.
+            memory: the memory to query; usually the output of an RNN encoder.
+                This tensor should be shaped ``[batch_size, max_time, ...]``.
+            memory_sequence_length (optional): sequence lengths for the batch
+                entries in memory.  If provided, the memory tensor rows are
+                masked with zeros for values past the respective sequence
+                lengths.
+        """
+        self._query = self.query_layer(query) if self.query_layer else query
```
huzecong: Is `self._query` cleared by the end of `forward`? This should also be cleared.

Actually, I would prefer making `query` the return value of the base class `forward` method, because we don't need to cache it, and storing it as an attribute gives more confusing `Optional` type annotations.

huzecong:
Great! I'll merge this PR once Zhiting sets up CI for this new repo.

PR #36, opened 2019-06-13
Connector

TomNong:
Added connector module.


texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,786 @@
...
+                batch size.
+            value (optional): A scalar, the value that the output tensor(s)
+                has. If ``None``, ``"value"`` in :attr:`hparams` is used.
+
+        Returns:
+            A (structure of) ``Tensor`` whose structure is the same as
+            :attr:`output_size`, with value speicified by
+            ``value`` or :attr:`hparams`.
+        """
+        value_ = self.hparams.value
```
huzecong: The reason we removed `value` from the argument is because we can infer the module device from the value in HParams. This requires us to convert value to a tensor in `__init__`, and registering it as a buffer of the module.

texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,786 @@
...
+
+        Returns:
+            A ``Tensor`` or a (nested) ``tuple`` of Tensors of the
+            same structure of :attr:`output_size`.
+        """
+        activation_fn = layers.get_activation_fn(self.hparams.activation_fn)
+
+        output, linear_layer = _mlp_transform(
+            inputs, self._output_size, activation_fn)
+        self._linear_layers.append(linear_layer)
```
huzecong: This is incorrect in several ways:
1. Submodules and parameters must be added in `__init__` method, because users can call `.to` after constructing models to move module parameters to appropriate devices.
2. Current impl. will result in a new linear layer to be created every time `forward` is called. This is clearly incorrect. Constructor of `MLPTransformConnector` should take the final dimension size of input tensors as an argument.

texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,786 @@
...
+        }
+
+    def forward(self, # type: ignore
+                transform: bool = True,
+                num_samples: Optional[Union[int, torch.Tensor]] = None
+                ) -> Tuple[Any, Any]:
+        r"""Samples from a distribution and optionally performs transformation
+        with an MLP layer.
+        The distribution must be reparameterizable, i.e.,
+        :python:`distribution.reparameterization_type = FULLY_REPARAMETERIZED`.
```
huzecong: PyTorch distributions do not have such attributes. In PyTorch, whether a distribution is reparameterizable can be inferred by accessing an undocumented class attribute: `Distribution.has_rsample`.

texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,786 @@
...
+
+    def __init__(self,
+                 output_size: OutputSize,
+                 hparams: HParamsType = None,
+                 distribution: Union[Type[T], T, str] = 'MultivariateNormal',
+                 distribution_kwargs: Optional[Dict[str, Any]] = None,):
+        ConnectorBase.__init__(self, output_size, hparams)
+        if distribution_kwargs is None:
+            distribution_kwargs = {}
+        self._dstr = utils.check_or_get_instance(
```
huzecong: Reparameterized distributions work in a different way in PyTorch. For those distributions, it is required to pass the parameters as arguments in the constructor (see `MultivariateNormal` for instance: it is required to pass `loc` (mean), and some representation of variance). Thus, the parameters must be instantiated in this `__init__` method, and stored as parameters of this module.

I checked all reparameterizable distributions in PyTorch and their constructors are pretty different. The required parameters have different names (and some has constraints on the initial values, e.g. has to be positive or positive definite), and there is no way to programmatically know the parameter names or required shapes. We may need to rethink how this module should be implemented.

One idea I have now is: Add support for automatic construction of the most common distributions (`MultivariateNormal`, and maybe a couple of others?), and for the rest, ask the user to manually add required arguments in `distribution_kwargs`. We might not need the user to construct the tensor for us, but rather have them specify the size and init_fn.

texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,786 @@
...
+            - sample: The sample from the distribution, prior to transformation.
+        Raises:
+            ValueError: If distribution cannot be reparametrized.
+            ValueError: The output does not match :attr:`output_size`.
+        """
+
+        if num_samples:
+            sample = self._dstr.sample([num_samples]) # type: ignore
+        else:
+            sample = self._dstr.sample() # type: ignore
```
huzecong: Also, we should use `rsample` instead of `sample`.

texar/modules/connectors/connectors.py
huzecong: Or, have the user manually construct the distribution. After that, we can extract the tensors within the distribution with some dirty hacks, wrap them as `nn.Parameter`s, register them as parameters of this module, and replace them back.

texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,786 @@
...
+            output, linear_layer = _mlp_transform(
+                sample, self._output_size, activation_fn)
+            self._linear_layers.append(linear_layer)
+
+            _assert_same_size(output, self._output_size)
+
+        return (output, sample)
+
+
+class StochasticConnector(ConnectorBase):
```
huzecong: This class has the same problems as above.

texar/modules/connectors/connectors_test.py
```diff
@@ -0,0 +1,200 @@
+#
+"""
+Unit tests for connectors.
```
huzecong: Could you add test cases where the output size is a tuple of `int`, and cases where it is a tuple with mixed `torch.Size` and `int` (not sure if we intend to support this)?

texar/modules/connectors/connectors.py
huzecong: PS. A buffer of a `nn.Module` is a tensor that is part of the module state but not a parameter. Buffers are moved along with parameters with `.cuda()` or `.to(device=)` is called. Please see <https://pytorch.org/docs/stable/nn.html#torch.nn.Module.register_buffer>

texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,786 @@
...
+]
+
+T = TypeVar('T')
+Inputs = Union[List[Any], Dict[Any, Any], Tuple[Any], torch.Tensor]
+D = MultivariateNormal
+OutputSize = Union[MaybeTuple[int], MaybeTuple[torch.Size], Tuple]
+HParamsType = Optional[HParams]
+#Optional[Union[HParams, dict]]
+ActivationFn = Optional[Callable[[torch.Tensor], torch.Tensor]]
+LinearLayer = Callable[[torch.Tensor], torch.Tensor]
```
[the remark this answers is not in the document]
TomNong: Got it! This part is fixed now.

texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,786 @@
...
+            (i.e., batch-major), or a (nested) tuple of such elements.
+            A Tensor or a (nested) tuple of Tensors with shape
+            ``[max_time, batch_size, ...]`` (i.e., time-major) can
+            be transposed to batch-major using
+            :func:`~texar.utils.transpose_batch_time` prior to this function.
+        output_size: Can be an ``Integer``, a ``torch.Size``, or a (nested)
+            ``tuple`` of ``Integers`` or ``torch.Size``.
+        activation_fn: Activation function applied to the output.
+
+    Returns:
```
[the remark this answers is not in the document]
TomNong: New to know this problem, checked!

texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,786 @@
...
+        returns a ``Tensor`` of shape ``[batch_size x output_size]``.
+        If :attr:`output_size` is a ``tuple`` of Integers or TensorShape,
+        returns a ``tuple`` having the same structure as:attr:`output_size`,
+        where each element ``Tensor`` has the same size as
+        defined in :attr:`output_size`.
+    """
+    # Flatten inputs
+    flat_input = nest.flatten(inputs)
+    dims = [_get_tensor_depth(x) for x in flat_input]
+    flat_input = [torch.reshape(x, (-1, d)) for x, d in zip(flat_input, dims)]
```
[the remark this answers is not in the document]
TomNong: That is much neater.

texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,786 @@
...
+            size_list[i] = np.prod([dim for dim in shape])
+    else:
+        size_list = flat_output_size
+
+    sum_output_size = sum(size_list)
+    linear_layer = nn.Linear(
+        concat_input.size()[-1], sum_output_size)
+
+    fc_output = linear_layer(concat_input)
+    if activation_fn:
```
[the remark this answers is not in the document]
TomNong: Checked

texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,786 @@
...
+        fc_output = activation_fn(fc_output)
+
+    flat_output = split(fc_output, size_list, dim=1) # type: ignore
```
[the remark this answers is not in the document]
TomNong: This one is checked too.

texar/modules/connectors/connectors.py
TomNong: Thanks! I know how to it next time if we want to infer the device etc.

texar/modules/connectors/connectors.py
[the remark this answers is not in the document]
TomNong: For the final_dim, I think we could made it an input from the user

texar/modules/connectors/connectors.py
TomNong: Thanks! That is not a easy-to-find one.

texar/modules/connectors/connectors.py
TomNong: I would say the second method is better, we can't cover all the reparameterizable distributions very uniformly.

texar/modules/connectors/connectors.py
TomNong: Thanks! Checked.

texar/modules/connectors/connectors.py
TomNong: Checked.

texar/modules/connectors/connectors_test.py
TomNong: Yeah actually I don't think the code should support mixture of types as output size, if user wants to use `torch.Size` and `int`, they may wrap `int` into `torch.Size`.

PR #49, opened 2019-06-19
Connector

TomNong:
1. Modified params and made linear layer attributes of the module.
2. Modified reparamatable distribution modules, re-deployed the distribution attributes.
3. Modified docs.


texar/modules/connectors/connectors.py
```diff
@@ -470,30 +502,47 @@ class ReparameterizedStochasticConnector(ConnectorBase):
...
-        self._dstr = utils.check_or_get_instance(
-            distribution, distribution_kwargs,
-            ["torch.distributions", "torch.distributions.multivariate_normal",
-             "texar.custom"])
+
+        self._dstr = distribution
+        for dstr_attr in self._dstr.arg_constraints.keys():  # type: ignore
+            tensor = getattr(self._dstr, dstr_attr)
+            self.register_buffer(dstr_attr, nn.Parameter(tensor))
+            setattr(self._dstr, dstr_attr, getattr(self, dstr_attr))
```
huzecong: Three mistakes here:
1. A parameter should not be registered as a buffer. Use `register_parameter` instead.
2. `register_*` stores the buffer as an attribute internally, so there's no need to call `setattr` here.
3. Implementations of the PyTorch distributions are more convoluted than I thought. To name a few:
   - Actual stored arguments might be different to what's stored inside `arg_constraints`. For instance, `Chi2` inherits from `Gamma` and has argument `df`, however the actual stored attributes are `concentration` and `rate`, which are arguments of the constructor of `Gamma`. 
   - `MultivariateNormal` has four arguments `loc`, `scale_tril`, `covariance_matrix`, and `precision_matrix`, but exactly one among the latter three should be provided. Internally everything is converted to a `scale_tril` matrix, but if the user chooses to parameterize `covariance_matrix` instead of `scale_tril`, we should not make `scale_tril` a parameter.

Sorry for not noticing these details before. Given such absurd implementation of PyTorch distributions, I think the best way to implement this Connector module is to manually support a couple of most commonly used distributions.

To still support arbitrary distributions, I think what we can do is to ask the user to provide the distribution type, arguments to fill in, and initial tensor values for those arguments. Then, we construct the distribution every time `forward` is called.

texar/modules/connectors/connectors.py
```diff
@@ -591,30 +624,49 @@ class StochasticConnector(ConnectorBase):
...
 
     def __init__(self,
                  output_size: OutputSize,
-                 hparams: HParamsType = None,
-                 distribution: Union[Type[T], T, str] = 'MultivariateNormal',
-                 distribution_kwargs: Optional[Dict[str, Any]] = None,):
+                 distribution: T,
+                 num_samples: Optional[Union[int, torch.Tensor]] = None,
+                 hparams: Optional[HParamsType] = None):
         ConnectorBase.__init__(self, output_size, hparams)
```
huzecong: Could you change calls to the super class constructor to this:
```python
super().__init__(output_size, hparams)
```
This is the new convention in Python, and is less error-prone.

texar/modules/connectors/connectors.py
```diff
@@ -41,19 +39,20 @@
...
-Inputs = Union[List[Any], Dict[Any, Any], Tuple[Any], torch.Tensor]
-D = MultivariateNormal
-OutputSize = Union[MaybeTuple[int], MaybeTuple[torch.Size], Tuple]
-HParamsType = Optional[HParams]
-#Optional[Union[HParams, dict]]
+TensorStruct = Union[List[torch.Tensor],
+                     Dict[Any, torch.Tensor],
+                     MaybeTuple[torch.Tensor]]
+OutputSize = MaybeTuple[Union[int, torch.Size]]
+HParamsType = HParams
```
huzecong: I don't think it's necessary to redefine `HParams` into a longer name...

texar/modules/connectors/connectors.py
```diff
@@ -84,15 +84,36 @@ def _get_tensor_depth(x: torch.Tensor) -> int:
...
+    Args:
+        output_size: Can be an ``Integer``, a ``torch.Size``, or a (nested)
+            ``tuple`` of ``Integers`` or ``torch.Size``.
+    """
+    flat_output_size = nest.flatten(output_size)
+
+    if isinstance(flat_output_size[0], torch.Size):
+        size_list = [0] * len(flat_output_size)
+        for (i, shape) in enumerate(flat_output_size):
+            size_list[i] = np.prod([dim for dim in shape])
```
huzecong: This is a very un-Pythonic implementation. Readability could be greatly improved by writing:
```python
size_list = [np.prod(shape) for shape in flat_output_size]
```

texar/modules/connectors/connectors.py
```diff
@@ -438,29 +467,32 @@ class ReparameterizedStochasticConnector(ConnectorBase):
...
+                cell.state_size, distribution=gauss_ds)
+            output, sample = connector()
+            # output == LSTMStateTuple(c=tensor_of_shape_(1, 256),
+            #                          h=tensor_of_shape_(1, 256))
+            # sample == Tensor([16 * 100])
+
+            # Initialized with num_samples
+            connector = ReparameterizedStochasticConnector(
+                cell.state_size, distribution=gauss_ds, num_samples=4)
             output_, sample_ = connector(distribution_kwargs=kwargs,
```
huzecong: Please update this code example. I don't think we have `distribution_kwargs` anymore.

PR #53, opened 2019-06-19
Add ScalarData and MultiAlignedData

AvinashBukkittu:
This PR adds the following

1) `ScalarData` with test cases
2) `MultiAlignedData` with test cases
3) Docs for above databases in `data.rst`

### DIFF
--- a/texar/data/data/mono_text_data.py
+++ b/texar/data/data/mono_text_data.py
@@ -75,7 +75,7 @@ class MonoTextData(TextDataBase[str, List[str]]):
             hyperparameters. See :meth:`default_hparams` for the defaults.
 
     By default, the processor reads raw data files, performs tokenization,
-    batching and other pre-processing steps, and results in a TF Dataset
+    batching and other pre-processing steps, and results in a Dataset
     whose element is a python `dict` including three fields:
 
     "text":
--- a/texar/data/data/mono_text_data.py
+++ b/texar/data/data/mono_text_data.py
@@ -22,10 +22,14 @@ import torch
 
 from texar.data.data.dataset_utils import Batch, padded_batch
 from texar.data.data.text_data_base import TextDataBase, TextLineDataSource
+from texar.data.data.data_base import SequenceDataSource
 from texar.data.embedding import Embedding
 from texar.data.vocabulary import SpecialTokens, Vocab
 from texar.hyperparams import HParams
 from texar.utils import utils
+
+# pylint: disable=protected-access
+
 
 __all__ = [
     "_default_mono_text_dataset_hparams",
@@ -161,6 +165,59 @@ class MonoTextData(TextDataBase[str, List[str]]):
                 compression_type=self._hparams.dataset.compression_type)
 
         super().__init__(data_source, hparams, device=device)
+
+    @classmethod
+    def _construct(cls, hparams, device: Optional[torch.device] = None,
+                   vocab: Optional[Vocab] = None,
+                   embedding: Optional[Vocab] = None):
+        mono_text_data = cls.__new__(cls)
+        mono_text_data._hparams = HParams(hparams,
+                                          mono_text_data.default_hparams())
+        if mono_text_data._hparams.dataset.variable_utterance:
+            raise NotImplementedError
+
+        dataset = mono_text_data._hparams.dataset
+        mono_text_data._other_transforms = dataset.other_transformations
+
+        # Create vocabulary
+        if vocab is not None:
+            mono_text_data._vocab = vocab
+            mono_text_data._bos_token = vocab.bos_token
+            mono_text_data._eos_token = vocab.eos_token
+        else:
+            mono_text_data._bos_token = dataset.bos_token
+            mono_text_data._eos_token = dataset.eos_token
+            bos = utils.default_str(mono_text_data._bos_token,
+                                    SpecialTokens.BOS)
+            eos = utils.default_str(mono_text_data._eos_token,
+                                    SpecialTokens.EOS)
+            mono_text_data._vocab = Vocab(dataset.vocab_file,
+                                          bos_token=bos, eos_token=eos)
+
+        # Create embedding
+        if embedding is not None:
+            mono_text_data._embedding = embedding
+        else:
+            mono_text_data._embedding = mono_text_data.make_embedding(
+                dataset.embedding_init,
+                mono_text_data._vocab.token_to_id_map_py)
+
+        mono_text_data._delimiter = dataset.delimiter
+        mono_text_data._max_seq_length = dataset.max_seq_length
+        mono_text_data._length_filter_mode = _LengthFilterMode(
+            mono_text_data._hparams.dataset.length_filter_mode)
+        mono_text_data._pad_length = mono_text_data._max_seq_length
+        if mono_text_data._pad_length is not None:
+            mono_text_data._pad_length += sum(int(x != '') for x in
+                                              [mono_text_data._bos_token,
+                                              mono_text_data._eos_token])
+
+        data_source: SequenceDataSource[str] = SequenceDataSource([])
+        super(MonoTextData, mono_text_data).__init__(source=data_source,
+                                                     hparams=hparams,
+                                                     device=device)
+
+        return mono_text_data
 
     @staticmethod
     def default_hparams():
@@ -354,10 +411,10 @@ class MonoTextData(TextDataBase[str, List[str]]):
             if self._length_filter_mode is _LengthFilterMode.TRUNC:
                 words = words[:self._max_seq_length]
 
-        if self._bos_token != '':
-            words.insert(0, self._bos_token)
-        if self._eos_token != '':
-            words.append(self._eos_token)
+        if self._hparams.dataset["bos_token"] != '':
+            words.insert(0, self._hparams.dataset["bos_token"])
+        if self._hparams.dataset["eos_token"] != '':
+            words.append(self._hparams.dataset["eos_token"])
 
         # Apply the "other transformations".
         for transform in self._other_transforms:
@@ -383,8 +440,9 @@ class MonoTextData(TextDataBase[str, List[str]]):
 
         text_ids = torch.from_numpy(text_ids).to(device=self.device)
         lengths = torch.tensor(lengths, dtype=torch.long, device=self.device)
-        return Batch(len(examples), text=examples,
-                     text_ids=text_ids, length=lengths)
+        batch = {self.text_name: examples, self.text_id_name: text_ids,
+                 self.length_name: lengths}
+        return Batch(len(examples), batch=batch)
 
     def list_items(self) -> List[str]:
         r"""Returns the list of item names that the data can produce.
@@ -404,6 +462,7 @@ class MonoTextData(TextDataBase[str, List[str]]):
         """
         return self._vocab
 
+    @property
     def text_name(self):
         r"""The name for the text field"""
         if self.hparams.dataset["data_name"]:
--- a/texar/data/data/mono_text_data.py
+++ b/texar/data/data/mono_text_data.py
@@ -347,7 +347,7 @@ class MonoTextData(TextDataBase[str, List[str]]):
         return embedding
 
     def process(self, raw_example: str) -> List[str]:
-        # `_process` truncates sentences and appends BOS/EOS tokens.
+        # Truncates sentences and appends BOS/EOS tokens.
         words = raw_example.split(self._delimiter)
         if (self._max_seq_length is not None and
                 len(words) > self._max_seq_length):
--- a/texar/data/data/multi_aligned_data.py
+++ b/texar/data/data/multi_aligned_data.py
@@ -68,7 +68,7 @@ def _is_scalar_data(data_type):
     return data_type == _DataTypes.INT or data_type == _DataTypes.FLOAT
 
 
-def _is_tfrecord_data(data_type):
+def _is_record_data(data_type):
     return data_type == _DataTypes.RECORD
 
 
@@ -87,7 +87,7 @@ def _default_dataset_hparams(data_type=None):
         })
     elif _is_scalar_data(data_type):
         hparams = _default_scalar_dataset_hparams()
-    elif _is_tfrecord_data(data_type):
+    elif _is_record_data(data_type):
         hparams = _default_record_dataset_hparams()
         hparams.update({
             "data_type": _DataTypes.RECORD,
@@ -105,9 +105,9 @@ class MultiAlignedData(
             defaults.
 
     The processor can read any number of parallel fields as specified in
-    the "datasets" list of :attr:`hparams`, and result in a TF Dataset whose
+    the "datasets" list of :attr:`hparams`, and result in a Dataset whose
     element is a python `dict` containing data fields from each of the
-    specified datasets. Fields from a text dataset or TFRecord dataset have
+    specified datasets. Fields from a text dataset or Record dataset have
     names prefixed by its "data_name". Fields from a scalar dataset are
     specified by its "data_name".
 
@@ -125,19 +125,18 @@ class MultiAlignedData(
             }
             data = MultiAlignedData(hparams)
             iterator = DataIterator(data)
-            batch = iterator.get_next()
 
-            iterator.switch_to_dataset(sess) # initializes the dataset
-            batch_ = sess.run(batch)
-            # batch_ == {
-            #    'x_text': [['<BOS>', 'x', 'sequence', '<EOS>']],
-            #    'x_text_ids': [['1', '5', '10', '2']],
-            #    'x_length': [4]
-            #    'y_text': [['<BOS>', 'y', 'sequence', '1', '<EOS>']],
-            #    'y_text_ids': [['1', '6', '10', '20', '2']],
-            #    'y_length': [5],
-            #    'z': [1000],
-            # }
+            for batch in iterator:
+                # batch contains the following
+                # batch_ == {
+                #    'x_text': [['<BOS>', 'x', 'sequence', '<EOS>']],
+                #    'x_text_ids': [['1', '5', '10', '2']],
+                #    'x_length': [4]
+                #    'y_text': [['<BOS>', 'y', 'sequence', '1', '<EOS>']],
+                #    'y_text_ids': [['1', '6', '10', '20', '2']],
+                #    'y_length': [5],
+                #    'z': [1000],
+                # }
 
             ...
 
@@ -162,20 +161,18 @@ class MultiAlignedData(
             }
             data = MultiAlignedData(hparams)
             iterator = DataIterator(data)
-            batch = iterator.get_next()
-
-            iterator.switch_to_dataset(sess) # initializes the dataset
-            batch_ = sess.run(batch)
-            # batch_ == {
-            #    'x_text': [['<BOS>', 'NewYork', 'City', 'Map', '<EOS>']],
-            #    'x_text_ids': [['1', '100', '80', '65', '2']],
-            #    'x_length': [5],
-            #
-            #    # "t_image" is a list of a "numpy.ndarray" image
-            #    # in this example. Its width equals to 512 and
-            #    # its height equals to 512.
-            #    't_image': [...]
-            # }
+            for batch in iterator:
+                # batch contains the following
+                # batch_ == {
+                #    'x_text': [['<BOS>', 'NewYork', 'City', 'Map', '<EOS>']],
+                #    'x_text_ids': [['1', '100', '80', '65', '2']],
+                #    'x_length': [5],
+                #
+                #    # "t_image" is a list of a "numpy.ndarray" image
+                #    # in this example. Its width equals to 512 and
+                #    # its height equals to 512.
+                #    't_image': [...]
+                # }
 
     """
 
@@ -211,7 +208,7 @@ class MultiAlignedData(
                                              hparams_i.compression_type)
                 datasources.append(datasource_i)
                 self._dataset_features.append(None)
-            elif _is_tfrecord_data(dtype):
+            elif _is_record_data(dtype):
                 datasource_i = PickleDataSource(file_paths=hparams_i.files)
                 datasources.append(datasource_i)
                 feature_types = hparams_i.feature_original_types
@@ -278,7 +275,7 @@ class MultiAlignedData(
 
     @staticmethod
     def default_hparams():
-        """Returns a dicitionary of default hyperparameters.
+        """Returns a dictionary of default hyperparameters.
 
         .. code-block:: python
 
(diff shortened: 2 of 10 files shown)


texar/data/data/multi_aligned_data.py
```diff
@@ -0,0 +1,701 @@
...
+        processed_examples = []
+        for i, raw_example_i in enumerate(raw_example):
+            if _is_text_data(dataset_hparams[i]["data_type"]):
+                delimiter = dataset_hparams[i]["delimiter"]
+                raw_tokens = raw_example_i.split(delimiter)  # type: ignore
+                max_seq_length = dataset_hparams[i]["max_seq_length"]
+                if (max_seq_length is not None and
+                        len(raw_tokens) > max_seq_length):
+                    if dataset_hparams[i]["length_filter_mode"] is \
+                            _LengthFilterMode.TRUNC.value:
```
huzecong: The `is` comparison only works when `dataset_hparams[i]["length_filter_mode"]` is a *string literal*, i.e. when the user says `"trunc"` in the code, but not in a config file that is read. Generally, use `==` for comparison unless you're sure comparing by `id` (pointer address) is fine.

texar/data/data/multi_aligned_data.py
```diff
@@ -0,0 +1,701 @@
...
+                if (dataset_hparams[i].eos_token is not None and
+                        dataset_hparams[i].eos_token != ''):
+                    raw_tokens.append(dataset_hparams[i].eos_token)
+                transforms = self._other_transforms[i]
+                if len(transforms) > 0:
+                    for transform in transforms:
+                        raw_tokens = transform(raw_tokens)
+                processed_examples.append(raw_tokens)
+            elif _is_scalar_data(dataset_hparams[i]["data_type"]):
+                scalar_ex: Union[str, int, float] = str(raw_example_i)
```
huzecong: Why is it necessary to convert it to `str` first?

texar/data/data/multi_aligned_data.py
```diff
@@ -0,0 +1,701 @@
...
+                        raw_tokens = raw_tokens[:max_seq_length]
+
+                if (dataset_hparams[i].bos_token is not None and
+                        dataset_hparams[i].bos_token != ''):
+                    raw_tokens.insert(0, dataset_hparams[i].bos_token)
+                if (dataset_hparams[i].eos_token is not None and
+                        dataset_hparams[i].eos_token != ''):
+                    raw_tokens.append(dataset_hparams[i].eos_token)
+                transforms = self._other_transforms[i]
+                if len(transforms) > 0:
```
huzecong: The `transform` logic is the same for all data types. Consider moving it out of the if branches. The `if len(transforms) > 0` check is also unnecessary.

texar/data/data/multi_aligned_data.py
```diff
@@ -0,0 +1,701 @@
...
+                    for transform in transforms:
+                        scalar_ex = transform(scalar_ex)
+                processed_examples.append(scalar_ex)
+            elif _is_record_data(dataset_hparams[i]["data_type"]):
+                example: Dict[str, Any] = raw_example_i  # type: ignore
+                # for some reasons even after the not None check here, mypy
+                # thinks self._dataset_features[i] may be None, hence adding
+                # ignore to the lines below
```
huzecong: I think this is because `self._dataset_features[i]` calls `__getitem__` of `_dataset_features`, which could be a user-defined function that modifies stuff. mypy can only infer that `__getitem__` of `List[T]` returns `T`, and nothing else.

If you store `self._dataset_features[i]` in a local variable and do a `is not None` check on it, ignores might not be needed anymore.

texar/data/data/multi_aligned_data.py
```diff
@@ -0,0 +1,701 @@
...
+
+        self._vocab = self.make_vocab(self._hparams.datasets)
+        self._embedding = self.make_embedding(self._hparams.datasets,
+                                              self._vocab)
+
+        self._dataset_features: List[Optional[Dict[str, Any]]] = []
+        self._name_prefix: List[str] = []
+        datasources: List[DataSource] = []
+        filters: List[Optional[Callable[[str], bool]]] = []
+        for _, hparams_i in enumerate(self._hparams.datasets):
```
huzecong: The logic below (and those in `process` and `collate`) have a great deal of duplicate code with the composed dataset classes (`MonoTextData`, `ScalarData`, `RecordData`). I guess the reason we need to copy their logic instead of using instances of those datasets are two fold:

1. `MonoTextData` does not provide interfaces to use shared vocabularies or embeddings.
2. Neither raw examples or processed examples can be directly accessed without constructing a `DataIterator` instance.

So my proposal is this:

- We create instances of the composed datasets as fully lazy and non-caching (so it does not access the data source nor store extra stuff), and use a null data source (e.g. `SequenceDataSource([])`).
- For `MonoTextData`, we add a private classmethod that constructs an instance of `MonoTextData` with possibly shared vocabulary & embeddings.
- In `process` and `collate`, we call the corresponding `process` and `collate` functions of each dataset.

texar/data/data/multi_aligned_data.py
```diff
@@ -0,0 +1,701 @@
...
+                        if (not torch.is_tensor(values) and
+                                descriptor.dtype not in [np.str_, np.bytes_]):
+                            values = torch.from_numpy(values)
+                    else:
+                        # VarLenFeature, just put everything in a Python list.
+                        pass
+                    batch.update({connect_name(self._name_prefix[i], key):
+                                  values})
+        return Batch(len(examples),
+                     **batch)
```
huzecong: Same here, no need to expand (or wrap the line).

texar/data/data/multi_aligned_data.py
```diff
@@ -0,0 +1,701 @@
...
+
+    @staticmethod
+    def _get_name_prefix(dataset_hparams):
```
huzecong: Where is this method used? This should be a private method but I don't see it used anywhere within the class.

texar/data/data/multi_aligned_data.py
```diff
@@ -0,0 +1,701 @@
...
+                    if len(sent) < pad_length else sent
+                    for sent in transposed_example
+                )
+
+                text_ids = torch.from_numpy(text_ids).to(device=self.device)
+                lengths = torch.tensor(lengths, dtype=torch.long,
+                                       device=self.device)
+
+                batch.update(
+                    {connect_name(self._name_prefix[i], "text"):
```
huzecong: Can we preprocess these names? These names won't be changed and it's a bit of a waste to recompute it every time.

texar/data/data/multi_aligned_data.py
```diff
@@ -0,0 +1,701 @@
...
+                items.append(self._name_prefix[i])
+            elif _is_record_data(ds_hpms["data_type"]):
+                for feature_name in \
+                        self._dataset_features[i]["feature_types"].keys():
+                    items.append(connect_name(self._name_prefix[i],
+                                 feature_name))
+        return items
+
+    @property
+    def dataset(self):
```
huzecong: We can make this a property of `DataBase`.

huzecong:
Oh, by the way, please avoid commit messages like "Apply the following". This is because the lines that follow are not displayed by default (either on GitHub or in local repo by `git log --graph`), and such message is not descriptive. Instead, consider summarizing the changes, for instance, your first commit could be "Add ScalarData and related tests & docs". If you find it hard to summarize all the different changes, then maybe it's best to split the commit into smaller ones that could be summarized on their own.

texar/data/data/multi_aligned_data.py
AvinashBukkittu: To avoid the mypy error - Incompatible types in assignment (expression has type "Union[str, Dict[str, Any]]", variable has type "Union[str, int, float]"). mypy throws this because raw example is of type `Union[str, Dict[str, Any]]` and the variable is `Union[str, int, float]`

huzecong: But the actual type must be a string right?
I would suggest not assigning `raw_example_i` to `scalar_ex`, but rather just call `int` or `float` on `raw_example_i` and store in `scalar_ex`. This way the type of `scalar_ex` would be `Union[int, float]` which make more sense, and mypy probably won't give warnings.

AvinashBukkittu: But `raw_example_i` can be of type `Dict[str, Any]`, right? In that case, mypy will throw an error if `int` or `float` is used directly on `raw_example_i`. We can directly type cast `raw_example_i` and store it in `scalar_ex` but we may have to `# type: ignore` it I think.

texar/data/data/mono_text_data.py
```diff
@@ -354,10 +411,10 @@ def process(self, raw_example: str) -> List[str]:
...
                 words = words[:self._max_seq_length]
 
-        if self._bos_token != '':
-            words.insert(0, self._bos_token)
-        if self._eos_token != '':
-            words.append(self._eos_token)
+        if self._hparams.dataset["bos_token"] != '':
+            words.insert(0, self._hparams.dataset["bos_token"])
+        if self._hparams.dataset["eos_token"] != '':
+            words.append(self._hparams.dataset["eos_token"])
```
huzecong: I don't think these need to be changed, as `_bos_token` and `_eos_token` are also added as attributes in the `_construct` method.

texar/data/data/multi_aligned_data.py
```diff
@@ -267,102 +277,101 @@ def _get_filter(delimiter, max_seq_length):
...
+                "prefetch_buffer_size": 0,
+                "max_dataset_size": -1,
+                "seed": None,
+                "name": "multi_aligned_data",
+            }
+
+        Here:
+
+        1. "datasets" is a list of `dict` each of which specifies a
+        dataset which can be text, scalar or Record. The :attr:`"data_name"`
```
huzecong: What I meant was to indent here, to the position after `"1. "`.

texar/data/data/multi_aligned_data.py
```diff
@@ -267,102 +277,101 @@ def _get_filter(delimiter, max_seq_length):
...
+        field of each dataset is used as the name prefix of the data fields
+        from the respective dataset. The :attr:`"data_name"` field of each
+        dataset should not be the same.
+
+            i) For scalar dataset, the allowed hyperparameters and default
+            values are the same as the "dataset" field of
```
huzecong: And probably here as well, after `"i) "`, if it is rendered as a numbered list as well.

texar/data/data/multi_aligned_data.py
```diff
@@ -267,102 +277,101 @@ def _get_filter(delimiter, max_seq_length):
...
+            :meth:`texar.data.RecordData.default_hparams`. Note that
+            :attr:`"data_type"` must be explicitly specified ("record").
+
+            iii) For text dataset, the allowed hyperparameters and default
+            values are the same as the "dataset" filed of
+            :meth:`texar.data.MonoTextData.default_hparams`,
+            with several extra hyperparameters:
+
+                `"data_type"`: str
```
huzecong: And these should be indented to the same column as the main body of text in this list item.

texar/data/data/multi_aligned_data.py
```diff
@@ -475,166 +484,43 @@ def process(self, raw_example: Tuple[Union[str, Dict[str, Any]], ...]) \
...
-                    scalar_ex = int(scalar_ex)
-                elif data_type == "float":
-                    scalar_ex = float(scalar_ex)
-                transforms = self._other_transforms[i]
-                if len(transforms) > 0:
-                    for transform in transforms:
-                        scalar_ex = transform(scalar_ex)
-                processed_examples.append(scalar_ex)
+                example_i = self._databases[i].process(raw_example_i)
             elif _is_record_data(dataset_hparams[i]["data_type"]):
```
huzecong: The checks are redundant --- the branch bodies are exactly the same.

texar/data/data/multi_aligned_data.py
```diff
@@ -475,166 +484,43 @@ def process(self, raw_example: Tuple[Union[str, Dict[str, Any]], ...]) \
...
-                            self._dataset_features[i]["convert_types"].items():  # type: ignore # pylint: disable=line-too-long
-                        example[key] = np.asarray(example[key], dtype=dtype)
-                    for key, transform in \
-                            self._dataset_features[i]["image_transforms"].items():  # type: ignore # pylint: disable=line-too-long
-                        example[key] = transform(example[key])
-                for transform in self._other_transforms[i]:
-                    example = transform(example)
-                processed_examples.append(example)
+                example_i = self._databases[i].process(raw_example_i)
+            for transform in self._other_transforms[i]:
```
huzecong: I don't think we need to deal with `other_transforms` anymore. They should be dealt with by individual data bases.

texar/data/data/multi_aligned_data.py
```diff
@@ -475,166 +484,43 @@ def process(self, raw_example: Tuple[Union[str, Dict[str, Any]], ...]) \
...
+                transform(example_i)
+            processed_examples.append(example_i)
         return tuple(processed_examples)
 
     def collate(self, examples) -> Batch:
         dataset_hparams = self.hparams["datasets"]
-        transposed_examples: List[Tuple] = list(map(tuple, zip(*examples)))
+        transposed_examples: \
+            List[List[Union[List[str], int, float, Dict[str, Any]]]] = \
+            list(map(list, zip(*examples)))
```
huzecong: The outermost `list` is redundant -- we only iterate through it once, so it's fine to keep it as a `map` iterator.

texar/data/data/multi_aligned_data.py
```diff
@@ -475,166 +484,43 @@ def process(self, raw_example: Tuple[Union[str, Dict[str, Any]], ...]) \
...
-                        to(device=self.device)
-                else:
-                    raise ValueError(
-                        "Incorrect \'data_type\'. Currently \'int\' and "
-                        "\'float\' are supported. Received {}"
-                        .format(data_type))
-                batch.update({self._name_prefix[i]: example})
+                scalar_batch = self._databases[i].collate(transposed_example)
+                batch.update(scalar_batch._batch)
             elif _is_record_data(dataset_hparams[i]["data_type"]):
```
huzecong: Same here, redundant checks.

texar/data/data/multi_aligned_data.py
```diff
@@ -687,8 +573,7 @@ def text_id_name(self, name_or_id):
         i = self._maybe_name_to_id(name_or_id)
```
huzecong: Implementation for all these `*_name` functions should change, since we don't store `_name_prefix` anymore. `_names` should be used directly.

texar/data/data/multi_aligned_data.py
```diff
@@ -220,35 +217,48 @@ def _get_filter(delimiter, max_seq_length):
...
+                         "text_ids": connect_name(hparams_i["data_name"],
+                                                  "text_ids"),
+                         "length": connect_name(hparams_i["data_name"],
+                                                "length")})
+
+                    text_hparams = MonoTextData.default_hparams()
+                    text_hparams.update({"dataset": hparams_i.todict()})
+
+                    # MonoTextData does not have following keys
+                    for key in ["data_type", "vocab_share_with",
```
huzecong: Instead of hard-coding keys, why not just iterate over all keys in `default_hparams`, and only updating those that are valid?

texar/data/data/multi_aligned_data.py
```diff
@@ -195,22 +191,23 @@ def __init__(self, hparams, device: Optional[torch.device] = None):
...
+        self._names: List[Dict[str, Any]] = []
         datasources: List[DataSource] = []
         filters: List[Optional[Callable[[str], bool]]] = []
-        for _, hparams_i in enumerate(self._hparams.datasets):
+        self._databases: List[DataBase] = []
+        for idx, hparams_i in enumerate(self._hparams.datasets):
             dtype = hparams_i.data_type
             datasource_i: DataSource
+
             if _is_text_data(dtype) or _is_scalar_data(dtype):
```
huzecong: I would prefer separating the logic of text data and scalar data, since the only thing they share is the construction of the `TextLineDataSource`. Separating them would make the code flow more clear.

texar/data/data/multi_aligned_data.py
```diff
@@ -195,22 +191,23 @@ def __init__(self, hparams, device: Optional[torch.device] = None):
...
-                self._dataset_features.append(None)
-                if _is_text_data(dtype) and hparams_i["length_filter_mode"] is \
-                        _LengthFilterMode.DISCARD.value and \
-                        hparams_i["max_seq_length"] is not None:
+                if _is_text_data(dtype) and (hparams_i["length_filter_mode"] is
+                                             _LengthFilterMode.DISCARD.value) \
+                        and (hparams_i["max_seq_length"] is not None):
 
                     def _get_filter(delimiter, max_seq_length):
                         return lambda x: len(x.split(delimiter)) <= \
```
huzecong: The `x` in this function is the tuple of raw examples from all sources, but the check should only be applied on the `idx`-th element of the tuple.

texar/data/data/multi_aligned_data.py
```diff
@@ -220,35 +217,48 @@ def _get_filter(delimiter, max_seq_length):
...
+                    self._databases.append(
+                        MonoTextData._construct(hparams=text_hparams,
+                                                device=device,
+                                                vocab=self._vocab[idx],
+                                                embedding=self._embedding[idx]))
+                else:
+                    self._names.append({"label": hparams_i["data_name"]})
+                    scalar_hparams = ScalarData.default_hparams()
+                    scalar_hparams.update({"dataset": hparams_i.todict()})
+                    self._databases.append(ScalarData(hparams=scalar_hparams,
```
huzecong: What about the source of this `ScalarData`? Won't this data base load all data and process them internally?
Same applies to `RecordData`.

huzecong:
In future please mark comments as solved when you resolve comments.

texar/data/data/multi_aligned_data.py
```diff
@@ -0,0 +1,586 @@
...
+        for idx, hparams_i in enumerate(self._hparams.datasets):
+            dtype = hparams_i.data_type
+            datasource_i: DataSource
+
+            if _is_text_data(dtype) or _is_scalar_data(dtype):
+                datasource_i = TextLineDataSource(
+                    hparams_i.files,
+                    compression_type=hparams_i.compression_type)
+                datasources.append(datasource_i)
+                if _is_text_data(dtype) and (hparams_i["length_filter_mode"] is
```
huzecong: The `is` is not yet changed. It should be `==`.

texar/data/data/mono_text_data.py
AvinashBukkittu: These were changed due to the issue I discussed with you last week, where `self.vocab["bos_token"]` and `self.vocab["eos_token"]` will be different from `hparams.dataset["bos_token"]` and `hparams.dataset["eos_token"]` in some cases. I was discussing with you the inconsistency between the two variables.

texar/data/data/multi_aligned_data.py
AvinashBukkittu: Not really, right? The `filter_fn` will receive tuple of raw examples from all sources, the `x` in individual filters is not a tuple, it is a component of the tuple on which filters have to be applied directly.

texar/data/data/multi_aligned_data.py
AvinashBukkittu: Resolved in 462312cab452a7f1c3eacce634d65f1a50c2fbbc

texar/data/data/multi_aligned_data.py
AvinashBukkittu: Resolved this in 462312cab452a7f1c3eacce634d65f1a50c2fbbc. `name_prefix` is no more an instance variable of the class but a temporary variable in the constructor needed to construct `self._name_to_id` dict. We now use `self._names` in all the getters.

texar/data/data/multi_aligned_data.py
AvinashBukkittu: Resolved in 462312cab452a7f1c3eacce634d65f1a50c2fbbc

texar/data/data/multi_aligned_data.py
AvinashBukkittu: Resolved in 462312cab452a7f1c3eacce634d65f1a50c2fbbc

texar/data/data/multi_aligned_data.py
AvinashBukkittu: Resolved in 462312cab452a7f1c3eacce634d65f1a50c2fbbc

texar/data/data/multi_aligned_data.py
AvinashBukkittu: Done! [462312cab452a7f1c3eacce634d65f1a50c2fbbc]

texar/data/data/multi_aligned_data.py
AvinashBukkittu: Created a `_construct` method in `ScalarData` and `RecordData` to solve this.

texar/data/data/multi_aligned_data.py
AvinashBukkittu: Done! [462312cab452a7f1c3eacce634d65f1a50c2fbbc]

texar/data/data/multi_aligned_data.py
huzecong: Ah, I see, you're right. Sorry for reading the code incorrectly.

texar/data/data/mono_text_data.py
huzecong: Yes, but when `MonoTextData` is constructed using the normal `__init__` constructor, or is constructed by calling `_construct` without specifying the vocabulary, `_bos_token` is just `hparams.dataset.bos_token`, which is what we want here. What I didn't understand is why we're setting `_bos_token` to the token stored in the vocabulary, if `_construct` is called with vocabulary specified. I think it would be more clear if we unified the behaviors.

AvinashBukkittu: I think unifying the two sets of variables will produce consistent behavior.

huzecong: Correct me if I'm wrong, but I thought the "inconsistent behavior" we discussed was only between hparams and the vocabulary -- `bos_token` in hparams could be `''` but is always a non-empty string in vocabulary. This is irrelevant of whether vocabularies are shared or not, the `bos_token` using in `process` should always be from the hparams. Please let me know if I understood any part of this incorrectly.

AvinashBukkittu: Yes, you are right - the inconsistency is in the special tokens between `hparams` and the `vocabulary`. Previously, in `process` method of `MonoTextData`, `bos_token` was used from the vocab, I changed it be picked from `hparams` in this PR.  Having said that, I am inclined to think this inconsistency may cause some minor issues in the future.

huzecong: What I suggested here is to use `_bos_token` in `process`, and always make `_bos_token` the token from `hparams`. I agree that the inconsistency could cause troubles for the user, but I think it is inconsistent for a reason. For example, the input files might already contain some kind of BOS & EOS tokens, so there's no point in appending them in `process`, but it's required to map them in the vocabulary.

For now, I think the best thing to do is to document this behavior, and leave a note for future fixes.

PR #56, opened 2019-06-23
Update meta info

ZhitingHu:
* Configure documentation, versioning
* Update READMEs

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,6 @@
+version: 2
+
+python:
+  version: 3.6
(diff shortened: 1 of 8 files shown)


huzecong:
The example in README probably needs changes. For one, we don't have `iterator.get_next()` anymore. Also, it might be better to follow PyTorch conventions and move all modules to an individual class, instead of defining them inline.

ZhitingHu:
Fixed in PR https://github.com/asyml/texar-pytorch/pull/59

Issue #63, opened 2019-06-25
seq2seq_attn example polish


ZhitingHu:
- [x] https://github.com/asyml/texar-pytorch/blob/dc51de295da9e861d739bf40ee4902cf8669ca53/examples/seq2seq_attn/seq2seq_attn.py#L23 ---> `from texar.core import ...` : do not expose too many library internal details 

- [x] https://github.com/asyml/texar-pytorch/blob/dc51de295da9e861d739bf40ee4902cf8669ca53/examples/seq2seq_attn/seq2seq_attn.py#L85 ---> `helper_train = self.decoder.create_helper(decoding_strategy="train_greedy")`: add readibility

- [x] https://github.com/asyml/texar-pytorch/blob/dc51de295da9e861d739bf40ee4902cf8669ca53/examples/seq2seq_attn/seq2seq_attn.py#L107
Is it necessary to convert to `tensor`. Can this be done internally?

- [x] https://github.com/asyml/texar-pytorch/blob/dc51de295da9e861d739bf40ee4902cf8669ca53/examples/seq2seq_attn/seq2seq_attn.py#L111
Concat right after `enc_outputs` is obtained

- [x] https://github.com/asyml/texar-pytorch/blob/dc51de295da9e861d739bf40ee4902cf8669ca53/examples/seq2seq_attn/seq2seq_attn.py#L113 
This should be in config file

- [x] https://github.com/asyml/texar-pytorch/blob/dc51de295da9e861d739bf40ee4902cf8669ca53/examples/seq2seq_attn/seq2seq_attn.py#L128
`parameters()` vs `trainable_variables()` ?

- [x] https://github.com/asyml/texar-pytorch/blob/dc51de295da9e861d739bf40ee4902cf8669ca53/examples/seq2seq_attn/seq2seq_attn.py#L132-L133
it's weird to get `iterator` from `data_iterator`

- [x] https://github.com/asyml/texar-pytorch/blob/dc51de295da9e861d739bf40ee4902cf8669ca53/examples/seq2seq_attn/seq2seq_attn.py#L137
Is it common?

- [x] https://github.com/asyml/texar-pytorch/blob/dc51de295da9e861d739bf40ee4902cf8669ca53/examples/seq2seq_attn/seq2seq_attn.py#L155
Mode in `{train, eval, predict}` or `{train, val}`

- [x] Add beam-search

huzecong:
[...]

PR #64, opened 2019-06-25
Resolve #60 Seq2Seq Example with GPU Support

### LINKED ISSUE #60: Seq2Seq Example with GPU Support
Hello,

How can I run the Seq2Seq example with my GPU?

I already modified the training data to use the cuda device as well as the model:
```
device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu") 

train_data = tx.data.PairedTextData(hparams=config_data.train, device=device)
val_data = tx.data.PairedTextData(hparams=config_data.val, device=device)
test_data = tx.data.PairedTextData(hparams=config_data.test, device=device)

model = Seq2SeqAttn(train_data)
model.to(device)
```

gpengzhi:
#60

### DIFF
--- a/examples/seq2seq_attn/seq2seq_attn.py
+++ b/examples/seq2seq_attn/seq2seq_attn.py
@@ -104,7 +104,7 @@ class Seq2SeqAttn(ModuleBase):
                 decoding_strategy="infer_greedy",
                 embedding=self.target_embedder,
                 start_tokens=start_tokens,
-                end_token=torch.tensor(self.eos_token_id))
+                end_token=start_tokens.new_tensor(self.eos_token_id))
 
             infer_outputs, _, _ = self.decoder(
                 helper=helper_infer,
@@ -153,7 +153,7 @@ def main():
         refs, hypos = [], []
         for batch in iterator:
             infer_outputs = model(batch, mode="infer")
-            output_ids = infer_outputs.sample_id
+            output_ids = infer_outputs.sample_id.cpu()
             target_texts_ori = [text[1:] for text in batch['target_text']]
             target_texts = tx.utils.strip_special_tokens(
                 target_texts_ori, is_token_list=True)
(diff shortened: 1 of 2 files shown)


examples/seq2seq_attn/seq2seq_attn.py
```diff
@@ -104,7 +104,7 @@ def forward(self, batch, mode):
                 decoding_strategy="infer_greedy",
                 embedding=self.target_embedder,
                 start_tokens=start_tokens,
-                end_token=torch.tensor(self.eos_token_id))
+                end_token=start_tokens.new_tensor(self.eos_token_id))
```
huzecong: Let's change this to a Python int, per #63.

examples/seq2seq_attn/seq2seq_attn.py
```diff
@@ -100,11 +100,12 @@ def forward(self, batch, mode):
         else:
             start_tokens = torch.ones_like(batch['target_length']) * \
                            self.bos_token_id
+
             helper_infer = self.decoder.create_helper(
                 decoding_strategy="infer_greedy",
                 embedding=self.target_embedder,
                 start_tokens=start_tokens,
-                end_token=start_tokens.new_tensor(self.eos_token_id))
```
huzecong: I might be wrong, but isn't `self.eos_token_id` already a Python int? Could you test run this to make sure it works?

gpengzhi: `self.eos_token_id` is a `numpy.int64`. If we use `self.eos_token_id` directly, we will get an error here (https://github.com/asyml/texar-pytorch/blob/master/texar/modules/decoders/decoder_helpers.py#L247).

huzecong: I see, thanks for clarifying!

PR #65, opened 2019-06-25
Resolve #54 AttentionRNNDecoder param initial_state wrapped in AttentionWrapperState

### LINKED ISSUE #54: AttentionRNNDecoder param initial_state wrapped in AttentionWrapperState
Hi @gpengzhi , we noticed that it would be too heavy to wrap the AttentionRNNDecoder forward method parameter "initial_state" into AttentionWrapperState. The position is here: https://github.com/asyml/texar-pytorch/blob/master/texar/modules/decoders/rnn_decoders.py#L518. Would you consider making it as a regular tensor? Thanks!

gpengzhi:
#54

### DIFF
--- a/texar/core/attention_mechanism.py
+++ b/texar/core/attention_mechanism.py
@@ -31,7 +31,7 @@ from torch import nn
 from texar.core.attention_mechanism_utils import (
     maybe_mask_score, prepare_memory, safe_cumprod)
 from texar.module_base import ModuleBase
-from texar.utils.types import MaybeTuple
+from texar.utils.types import MaybeList, MaybeTuple
 
 __all__ = [
     "AttentionMechanism",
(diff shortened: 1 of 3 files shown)


texar/modules/decoders/rnn_decoders.py
```diff
@@ -486,6 +486,34 @@ def default_hparams():
...
+            if isinstance(initial_state, list):
+                if isinstance(initial_state[0], tuple):
+                    batch_size = initial_state[0][0].shape[0]
+                else:
+                    batch_size = initial_state[0].shape[0]
+            else:
+                if isinstance(initial_state, tuple):
+                    batch_size = initial_state[0].shape[0]
+                else:
+                    batch_size = initial_state.shape[0]
```
huzecong: I actually suggest using `map_structure` for this. It could be done as:
```python
batch_size = None

def _get_batch_size_fn(x):
	nonlocal batch_size
	if isinstance(x, torch.Tensor):
		batch_size = x.size(0)

map_structure(_get_batch_size_fn, initial_state)
```

texar/core/attention_mechanism.py
```diff
@@ -941,3 +941,38 @@ class AttentionWrapperState(NamedTuple):
     attention_state: MaybeTuple[torch.Tensor]
     r"""A single or tuple of nested objects containing attention mechanism
     states for each attention mechanism."""
+
+    def clone(self, **kwargs):
```
huzecong: And I don't think the `clone` method is really necessary. It's only used once and this check is really redundant since we create the initial `AttentionWrapperState` based on the size of the provided `initial_state`.

Issue #62, opened 2019-06-25
BERT Classifier: Loss is 0?


bigabig:
Hey,

I am trying to run the BERT Classifier example with the MRPC task. 

I get the following error:
```
Traceback (most recent call last):
  File "C:\Program Files\JetBrains\PyCharm 2018.3.3\helpers\pydev\pydevd.py", line 1741, in <module>
    main()
  File "C:\Program Files\JetBrains\PyCharm 2018.3.3\helpers\pydev\pydevd.py", line 1735, in main
    globals = debugger.run(setup['file'], None, None, is_module)
  File "C:\Program Files\JetBrains\PyCharm 2018.3.3\helpers\pydev\pydevd.py", line 1135, in run
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "C:\Program Files\JetBrains\PyCharm 2018.3.3\helpers\pydev\_pydev_imps\_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "C:/Development/Git/texar-pytorch/examples/bert/bert_classifier_main.py", line 274, in <module>
    main()
  File "C:/Development/Git/texar-pytorch/examples/bert/bert_classifier_main.py", line 258, in main
    _train_epoch()
  File "C:/Development/Git/texar-pytorch/examples/bert/bert_classifier_main.py", line 156, in _train_epoch
    input_ids = batch["input_ids"]
  File "C:\Development\Git\texar-pytorch\texar\data\data\dataset_utils.py", line 78, in __getitem__
    return self._batch[item]
KeyError: 'input_ids'
```

Changing the keys as for example in input_ids = batch["input_ids"] to input_ids = batch["data_input_ids"] seems to fix these errors.

However, then i tried the script like this 
`python bert_classifier_main.py --do_train --do_eval`

and get the following logs:
```
python bert_classifier_main.py --do_train --do_eval
Pretrained model loaded from bert_pretrained_models/uncased_L-12_H-768_A-12\bert_model.ckpt
INFO:root:step: 50; loss: 0
INFO:root:step: 100; loss: 0
INFO:root:step: 150; loss: 0
INFO:root:step: 200; loss: 0
INFO:root:step: 250; loss: 0
INFO:root:step: 300; loss: 0
INFO:root:eval accu: 0.8578; loss: 0.3904; nsamples: 408
```
Is this correct?

huzecong:
[...]

PR #68, opened 2019-06-25
Handle prepend logic to keys in MultiAlignedData's collate function

AvinashBukkittu:
This PR

- Avoids prepending `data_name` to keys in RecordData
- Handles prepend logic for keys inside `MultiAlignedData`'s `collate` function instead of in individual components' `collate` function

This PR should fix the `input_ids` KeyError mentioned in https://github.com/asyml/texar-pytorch/issues/62.

### DIFF
--- a/texar/data/data/multi_aligned_data.py
+++ b/texar/data/data/multi_aligned_data.py
@@ -223,6 +223,8 @@ class MultiAlignedData(
                 for key in text_hparams["dataset"].keys():
                     if key in hparams_i:
                         text_hparams["dataset"][key] = hparams_i[key]
+                # handle prepend logic in MultiAlignedData collate function
+                text_hparams["dataset"]["data_name"] = None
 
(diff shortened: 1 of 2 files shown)


texar/data/data/multi_aligned_data.py
```diff
@@ -488,7 +490,8 @@ def collate(self, examples) -> Batch:
         batch: Dict[str, Any] = {}
         for i, transposed_example in enumerate(transposed_examples):
             kth_batch = self._databases[i].collate(transposed_example)
-            batch.update(kth_batch._batch)
+            for key, _ in self._names[i].items():
+                batch.update({self._names[i][key]: kth_batch[key]})
```
huzecong: Since we're already iterating over `.items()`, why not just write:
```python
for key, name in self._names[i].items():
    batch.update({name: kth_batch[key]})
```

PR #73, opened 2019-06-26
Resolve #70 Util bleu_moses.py does not work on windows

### LINKED ISSUE #70: Util bleu_moses.py does not work on windows
Hey,

When for example executing the seq2seq attention example, the function `corpus_bleu_moses` in `bleu_moses.py ` is used to evaluate the current model. This does not work on windows:

```
Traceback (most recent call last):
  File "seq2seq_attn.py", line 190, in <module>
    main()
  File "seq2seq_attn.py", line 178, in main
    val_bleu = _eval_epoch('val')
  File "seq2seq_attn.py", line 172, in _eval_epoch
    hypotheses=hypos)
  File "c:\development\git\texar-pytorch\texar\evals\bleu_moses.py", line 154, in corpus_bleu_moses
    multi_bleu_cmd, stdin=hyp_input, stderr=subprocess.STDOUT)
  File "C:\Development\Python36\lib\subprocess.py", line 336, in check_output
    **kwargs).stdout
  File "C:\Development\Python36\lib\subprocess.py", line 403, in run
    with Popen(*popenargs, **kwargs) as process:
  File "C:\Development\Python36\lib\subprocess.py", line 709, in __init__
    restore_signals, start_new_session)
  File "C:\Development\Python36\lib\subprocess.py", line 997, in _execute_child
    startupinfo)
OSError: [WinError 193] %1 is not a valid Win32 application
```

I fixed this error by changing line 147 of bleu_moses.py to
`multi_bleu_cmd = ["perl"] + [multi_bleu_path]`

I am not sure if this works without "perl" on Linux. Also, I don't know if you want to support Windows at all. But maybe you could state somewhere in the documentation that perl is required for some things (but I guess its already pre-installed on Linux, right?).

gpengzhi:
#70

### DIFF
--- a/examples/seq2seq_attn/seq2seq_attn.py
+++ b/examples/seq2seq_attn/seq2seq_attn.py
@@ -164,8 +164,7 @@ def main():
                 hypos.append(hypo)
                 refs.append([ref])
 
-        return tx.evals.corpus_bleu_moses(list_of_references=refs,
-                                          hypotheses=hypos)
+        return tx.evals.corpus_bleu(list_of_references=refs, hypotheses=hypos)
 
     best_val_bleu = -1.
     for i in range(config_data.num_epochs):
--- a/texar/evals/bleu.py
+++ b/texar/evals/bleu.py
@@ -117,7 +117,7 @@ def corpus_bleu(list_of_references: List[List[MaybeList[str]]],
                 max_order: int = 4,
                 lowercase: bool = False,
                 smooth: bool = False,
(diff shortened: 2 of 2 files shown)


huzecong:
I don't really think this is necessary... The reason we have separate methods `corpus_bleu` and `corpus_bleu_moses` is because these methods may return slightly different BLEU values. Simply changing from one to another does not solve the problem.

If the problem is due to certain Windows environments not recognizing the shebang line, let's just explicitly call `perl` in the command line.

PR #79, opened 2019-06-26
Add unit test for beam search & Fix bug when using `stop_early`

gpengzhi:
The current implementation does not execute as expected when `stop_early=True.` Modify the `_is_finished` function by following the code structure of `beam_search` in `tensor2tensor`.


haoransh:
For the change of code format, I'm not sure which is better. I'm using black to automate such stuff but maybe it's not perfect. @huzecong Could you have a look and maybe we could discuss to decide?

For the copyright of tensor2tensor, I suggest you add a permalink (for example, https://github.com/tensorflow/tensor2tensor/blob/eb048f69c7ea860324122b87cb9caf59c52a27f3/tensor2tensor/utils/beam_search.py) instead of (https://github.com/tensorflow/tensor2tensor/blob/master/tensor2tensor/utils/beam_search.py). In this way, we can avoid the potential affect of tensor2tensor update.

huzecong:
Regarding code format, I think either is okay. Personally I don't agree with Black on certain formats, but overall I think it looks good. As long as the code passes checks by `flake8`, it's okay.

But when modifying existing code, it's best to not change the code format (except the parts you modified, but even for that part it's best to keep consistent). Making code format changes would introduce excessive code diff results, making reviewing difficult.

texar/utils/beam_search.py
```diff
@@ -154,7 +157,7 @@ def compute_topk_scores_and_seq(
         batch_size: int
         states_to_gather: (possibly nested structure of) decoding states.
 
-    :returns: Tuple of:
+    Returns: Tuple of:
```
huzecong: The reason we use `:returns:` instead of `Returns:` here is because the Sphinx `napoleon` extension incorrectly regards the text between the colons (`Tuple of`) as the return type. The only way to workaround this is to use the raw returns role `:returns:`.

texar/utils/beam_search.py
```diff
@@ -120,15 +123,15 @@ def compute_batch_indices(batch_size: int, beam_size: int) -> torch.LongTensor:
...
-def compute_topk_scores_and_seq(
-    sequences: torch.LongTensor,
-    scores: torch.Tensor,
-    scores_to_gather: torch.Tensor,
-    flags: torch.ByteTensor,
-    beam_size: int,
-    batch_size: int,
-    states_to_gather: Optional[State] = None,
-) -> Tuple[torch.LongTensor, torch.Tensor, torch.ByteTensor, Optional[State]]:
+def compute_topk_scores_and_seq(sequences: torch.LongTensor,
```
huzecong: Code format changes like this are unnecessary and creates irrelevant clutter in code diffs. Please consider reverting these changes.

texar/utils/beam_search.py
```diff
@@ -16,7 +16,10 @@
 # ==============================================================================
 """
 Implementation of beam search with penalties.
-Adapted from tensor2tensor repository.
+
+Adapted from:
+    `https://github.com/tensorflow/tensor2tensor/`
+    `blob/master/tensor2tensor/utils/beam_search.py`
```
huzecong: URLs should not be broken into multiple lines. They are skipped during line length checks and spelling checks.

huzecong: And I agree with @haoransh on adding a permalink.

Issue #44, opened 2019-06-17
Conv1DNetwork returns tensor of incorrect shape


huzecong:
When using `Conv1DNetwork` with multiple filters and no dense layers, the returned tensor has incorrect shape.

For example:
```python
char_embed_size = 8
char_cnn = tx.modules.Conv1DEncoder(
    in_channels=char_embed_size, hparams={
        "kernel_size": [3, 4, 5],
        "out_channels": 50,
        "num_dense_layers": 0,
        "conv_activation": tx.core.identity,
        "dropout_conv": [],
        "dropout_rate": 0.0,
    })
batch_size, seq_len = 20, 30
# (batch, in_channels, in_features)
input = torch.randn(batch_size, char_embed_size, seq_len)
output = char_cnn(input)
out_channels = 150  # out_channels is multiplied by length of kernel_size
assert output.size() == (batch_size, out_channels)  # raises AssertionError
# actual returned size is (batch_size, 50, 3)
```

This is because internally `Conv1DNetwork` constructs a `MergeLayer` that concats outputs from differently-sized kernels, but does not specify the `dim` argument, so the default value of 2 is used. However, for convolutional layers, the channel dimension (defaults to dimension 1 in PyTorch) should be concat'd.

The test cases failed to capture this bug because dense layers were used. This resulted in `Flatten` layer being added, and `in_features` to the following `Linear` layer is inferred by actually running an input example through the network.

PR #80, opened 2019-06-26
Fix MergeOp when using reduce pooling ops

AvinashBukkittu:
This PR

- Fixes a bug during merging of convolutional layers. Specifically, when reduce pooling operations are used, the feature dim is removed and hence merging happens along channel dim
- Adds a test case in `Conv1DNetworkTest` to verify the above behavior.
- Tests the output of `MergeLayer` for all modes

This fixes https://github.com/asyml/texar-pytorch/issues/44

### DIFF
--- a/texar/core/layers.py
+++ b/texar/core/layers.py
@@ -690,7 +690,7 @@ class MergeLayer(nn.Module):
         # In case of reduce pooling operations, feature dim is removed and
         # channel dim is merged.
         # In non-reduce pooling operations, feature dim is merged.
-        dim = self._dim if self._dim is not None else - 1
+        dim = self._dim if self._dim is not None else -1
 
         if self._mode == 'concat':
             outputs = torch.cat(tensors=layer_outputs, dim=dim)
--- a/texar/core/layers.py
+++ b/texar/core/layers.py
@@ -647,6 +647,24 @@ class MergeLayer(nn.Module):
                 else:
                     self._layers.append(get_layer(hparams=layer))
 
+        self._functions = {
+            "sum": torch.sum,
+            "mean": torch.mean,
+            "prod": torch.prod,
+            "max": torch.max,
+            "min": torch.min,
+            "and": torch.all,
+            "or": torch.any,
+            "logsumexp": torch.logsumexp
+        }
+
+        def merge(tensors, dim):
+            _concat = torch.cat(tensors=tensors, dim=dim)
+            outputs = self._functions[self._mode](_concat, dim=dim)
+            return outputs
+
+        self._merge_op = merge
+
     def forward(self, input: torch.Tensor) -> torch.Tensor:  # type: ignore
         r"""Feed input to every containing layer and merge the outputs.
 
@@ -656,8 +674,11 @@ class MergeLayer(nn.Module):
         Returns:
             The merged tensor.
         """
+        layer_outputs: List[torch.Tensor]
         if self._layers is None:
-            layer_outputs: Union[torch.Tensor, List[torch.Tensor]] = input
+            layer_outputs = input
+            if not isinstance(layer_outputs, (list, tuple)):
+                layer_outputs = [layer_outputs]
         else:
             layer_outputs = []
             for layer in self._layers:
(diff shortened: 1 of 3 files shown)


texar/core/layers.py
```diff
@@ -664,6 +664,14 @@ def forward(self, input: torch.Tensor) -> torch.Tensor:  # type: ignore
                 layer_output = layer(input)
                 layer_outputs.append(layer_output)
 
+        # the merge dimension cannot be determined until we get the output from
+        # individual layers.
+        # In case of reduce pooling operations, feature dim is removed and
+        # channel dim is merged.
+        # In non-reduce pooling operations, feature dim is merged.
+        if self._dim is None:
+            self._dim = layer_outputs[0].dim() - 1
```
huzecong: Prefer not to change `self._dim`. It might be possible for the user to pass in tensors of different ranks.

texar/core/layers.py
```diff
@@ -664,6 +664,14 @@ def forward(self, input: torch.Tensor) -> torch.Tensor:  # type: ignore
                 layer_output = layer(input)
```
huzecong: (This comment is for line 660 but somehow I can't add a comment there.)

If `self._layers` is `None`, is it valid for `input` to be a tensor? To me it seems more reasonable if `input` is a list of tensors. This way the behaviors between cases w/ and w/o `self._layers` are consistent. Also, `torch.cat` requires the argument to be a list of tensors.

texar/core/layers.py
```diff
@@ -664,6 +664,14 @@ def forward(self, input: torch.Tensor) -> torch.Tensor:  # type: ignore
...
+        # individual layers.
+        # In case of reduce pooling operations, feature dim is removed and
+        # channel dim is merged.
+        # In non-reduce pooling operations, feature dim is merged.
+        if self._dim is None:
+            self._dim = layer_outputs[0].dim() - 1
+
         if self._mode == 'concat':
             outputs = torch.cat(tensors=layer_outputs, dim=self._dim)
         elif self._mode == 'elemwise_sum':
```
huzecong: (This comment is for lines 685-708.)

The logic here is almost the same, maybe we can unify them by e.g. creating a class-level dictionary mapping modes to torch functions and just looking them up.

texar/core/layers.py
AvinashBukkittu: I considered that possibility. `dim` arguments to torch functions in `forward` depends on the size of the input tensor. I am thinking to remove `self._dim` and make it function variable. In that case, we should be not be taking `dim` argument in the constructor. Also, `dim` is not being passed with any value during construction of `MergeLayer`. I think we can safely remove this. However, if you consider `texar-TF` interface compatibility, we may have to keep this.

texar/core/layers.py
AvinashBukkittu: `input` is the input to conv network or output of previous layers of the conv network, so as far as I know `input` is a `torch.Tensor`. The `if self._layers is None` handles a corner case where we allow `MergeLayer` to have no layers, in which case, we just return the input tensor. What we can do is, if the `if` condition holds true, we can set `layer_outputs = [inputs]`, in that way, `layer_outputs` is always a list.

texar/core/layers.py
AvinashBukkittu: True, I agree with that. Removes a lot of repetition. This was copied with little to no changes from `texar-TF`.

texar/core/layers.py
huzecong: We don't really need to move the `dim` argument to `forward`. What I meant was:
```python
dim = self._dim is self._dim is not None else -1
```
and use `dim` in the merging ops.

texar/core/layers.py
huzecong: So when `self._layers` is `None`, `MergeLayer` acts as identity? I don't really think we need to handle this; we just don't construct the `MergeLayer` in that case, which saves an unnecessary function call during `forward`.

If we do need to support that case, then we should just return the `input` instead of executing the rest of the logic. And we should have a test case for that.

texar/core/layers.py
```diff
@@ -647,6 +647,24 @@ def __init__(self, layers: Optional[List[nn.Module]] = None,
...
+            "or": torch.any,
+            "logsumexp": torch.logsumexp
+        }
+
+        def merge(tensors, dim):
+            _concat = torch.cat(tensors=tensors, dim=dim)
+            outputs = self._functions[self._mode](_concat, dim=dim)
+            return outputs
+
+        self._merge_op = merge
```
huzecong: I don't think it's necessary to store a function as attribute. The logic is simple and can be expanded in `forward` directly.

texar/core/layers.py
AvinashBukkittu: Done!

texar/core/layers.py
AvinashBukkittu: Yes, it acts as an Identity operator. As discussed offline, we will provide this feature to maintain compatibility with `texar-TF`. I have added a test case to cover `self._layers = None` case.

texar/core/layers.py
AvinashBukkittu: Done!

Issue #67, opened 2019-06-25
gpt2 example polish


ZhitingHu:
- [x] Port [gpt2_train_main.py](https://github.com/asyml/texar/blob/master/examples/gpt-2/gpt2_train_main.py) too

- [x] https://github.com/asyml/texar-pytorch/blob/dc51de295da9e861d739bf40ee4902cf8669ca53/examples/gpt-2/gpt2_generate_main.py#L26-L28
do not expose internal details. ---> `from texar.modules import TransformerDecoder`

- [x] https://github.com/asyml/texar-pytorch/blob/dc51de295da9e861d739bf40ee4902cf8669ca53/examples/gpt-2/gpt2_generate_main.py#L84-L95
Too sparse. Reformat to sth like 
```
parser.add_argument(
    '--config_model',
    ...
```

- [x] https://github.com/asyml/texar-pytorch/blob/dc51de295da9e861d739bf40ee4902cf8669ca53/examples/gpt-2/gpt2_generate_main.py#L100
Should this inherit `nn.Module` or `texar.ModuleBase`?

- [x] https://github.com/asyml/texar-pytorch/blob/dc51de295da9e861d739bf40ee4902cf8669ca53/examples/gpt-2/gpt2_generate_main.py#L197-L198
Better to print a message

huzecong:
Reopening this issue because [gpt2_train_main.py](https://github.com/asyml/texar/blob/master/examples/gpt-2/gpt2_train_main.py) is not yet ported.

huzecong:
I don't think this issue has been fully addressed, at least not in the master branch.

gpengzhi:
I am working on `gpt` modules and can take care of the remaining work.

PR #82, opened 2019-06-27
Polish gpt2 example, solve issue #67

TomNong:
Polish the gpt example, modified its layout.

### DIFF
--- a/examples/gpt-2/gpt2_generate_main.py
+++ b/examples/gpt-2/gpt2_generate_main.py
@@ -29,70 +29,56 @@ from texar.modules.decoders.transformer_decoders import TransformerDecoder
 from utils import model_utils, processor
 
 parser = argparse.ArgumentParser()
-parser.add_argument('--checkpoint',
-                    type=str,
-                    default=None,
-                    help="Model checkpoint to load model weights from. Use "
-                         "`--pretrain_checkpoint` instead if loading OpenAI "
-                         "pretrained checkpoint.")
-parser.add_argument('--pretrain_checkpoint',
-                    type=str,
-                    default="gpt2_pretrained_models/model_117M/model.ckpt",
-                    help="OpenAI pretrained model checkpoint. Ignored if "
-                         "'--checkpoint' is specified.")
-parser.add_argument('--pretrain_model_dir',
-                    type=str,
-                    default="gpt2_pretrained_models/model_117M",
-                    help="The directory of pretrained model, for loading "
-                         "vocabuary, etc.")
-parser.add_argument('--seed',
-                    type=int,
-                    default=None,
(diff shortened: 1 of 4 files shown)


examples/gpt-2/gpt2_generate_main.py
```diff
@@ -23,81 +23,91 @@
...
-                    help="The model configuration file to configure the "
-                         "model. The config file type is define by the "
-                         "'config_type',it be of texar type or json type."
-                         "For '--config_type=json', set the json "
-                         "config file path like: '--config_model "
-                         "gpt2_pretrained_models/model_117M/hparams.json';"
-                         "For '--config_type=texar', set the texar "
-                         "config file like: "
-                         "'--config_model configs.config_model_117M'.")
+parser.add_argument(
```
huzecong: This doesn't really address the "sparsity" issue brought up by @ZhitingHu. The arguments still take up a lot of unnecessary space. For instance, the argument name, type, and default values could fit on the same line. The help string can also be restructured to take fewer lines.

examples/gpt-2/gpt2_generate_main.py
```diff
@@ -195,13 +205,16 @@ def run_model():
                     print('Input should not be empty!')
                     raw_text = input("Model input >>> ")
             except EOFError:
+                print("Input not supported.")
```
huzecong: The message is not accurate. `EORError` is raised when the user sends an `EOF` character as input, in Unix-based systems this is `Ctrl+D`. This indicates that the users wants to quit the interactive environment, so a response of `"Input not supported."` would be confusing.

Personally I don't think a message is needed here, for instance `python` does not print anything on EOF. Other interactive envs like `ipython` print a confirmation message `"Do you really want to exit ([y]/n)?"`, but I don't think it's necessary here. If we really want a message, let's just say something along the lines of `"EOF entered, quitting."`.

examples/gpt-2/gpt2_generate_main.py
```diff
@@ -23,81 +23,91 @@
...
+         "gpt2_pretrained_models/model_117M/hparams.json';"
+         "For '--config_type=texar', set the texar "
+         "config file like: "
+         "'--config_model configs.config_model_117M'.")
 
 args = parser.parse_args()
 
 
-class GPT2(nn.Module):
+class GPT2(tx.ModuleBase):
```
huzecong: The discussion from our meeting was that for classes without their own `default_hparams()` method, it's fine to just use `nn.Module`. `tx.ModuleBase` is more like an internal base class for Texar modules.

huzecong: This is not yet reverted. The PR can be merged after this is taken care of.

PR #84, opened 2019-06-28
fix some doc.

wwt17:
These doc errors may also occur in the Tensorflow version.


huzecong:
Thanks for the contribution! The Transformer example was refactored a bit in the latest commit (we moved the entire model to a separate file), so this PR must be merged with master. Could you merge locally and update the PR? Thanks!

PR #85, opened 2019-06-28
Resolve #63 seq2seq_attn example polish & Resolve #55 AttentionRNNDecoder interface refactor

### LINKED ISSUE #55: AttentionRNNDecoder interface refactor
Requiring `input_size` and `encoder_output_size`, especially in `AttentionRNNDecoder`, is too much, e.g., [here](https://github.com/asyml/texar-pytorch/blob/master/examples/seq2seq_attn/seq2seq_attn.py#L66-L69):
```
tx.modules.AttentionRNNDecoder(
            encoder_output_size=(self.encoder.cell_fw.hidden_size +
                                 self.encoder.cell_bw.hidden_size),
            input_size=self.target_embedder.dim + config_model.decoder
            ['attention']['attention_layer_size'],
            ...)
```

Can we refactor the interface a bit so that sth like
```
encoder_output_size=self.encoder.output_size,
input_size=self.target_embedder.dim
```
is enough?

related issue: https://github.com/asyml/texar-pytorch/issues/42

@gpengzhi @huzecong @TomNong

### LINKED ISSUE #63: seq2seq_attn example polish

gpengzhi:
#63 
1. add `beam_search` to `attn_decoder`
2. code refactor on `optimization`
3. polish `seq2seq_attn` exmaple
4. resolve #55

### DIFF
--- a/examples/seq2seq_attn/seq2seq_attn.py
+++ b/examples/seq2seq_attn/seq2seq_attn.py
@@ -21,7 +21,6 @@ import torch
 import torch.nn as nn
 
 import texar as tx
-from texar.core import optimization
 
 parser = argparse.ArgumentParser()
 parser.add_argument('--config_model',
@@ -67,17 +66,11 @@ class Seq2SeqAttn(nn.Module):
         self.decoder = tx.modules.AttentionRNNDecoder(
             encoder_output_size=(self.encoder.cell_fw.hidden_size +
                                  self.encoder.cell_bw.hidden_size),
-            input_size=self.target_embedder.dim + config_model.decoder
-            ['attention']['attention_layer_size'],
+            input_size=self.target_embedder.dim,
             vocab_size=self.target_vocab_size,
             hparams=config_model.decoder)
 
     def forward(self, batch, mode):
-
-        if mode == "train":
-            self.train()
-        else:
-            self.eval()
 
         enc_outputs, _ = self.encoder(
             inputs=self.source_embedder(batch['source_text_ids']),
@@ -103,8 +96,9 @@ class Seq2SeqAttn(nn.Module):
 
             return mle_loss
         else:
-            start_tokens = torch.ones_like(batch['target_length']) * \
-                           self.bos_token_id
+            start_tokens = memory.new_full(batch['target_length'].size(),
+                                           self.bos_token_id,
+                                           dtype=torch.int64)
 
             infer_outputs = self.decoder(
                 start_tokens=start_tokens,
@@ -131,11 +125,12 @@ def main():
 
     model = Seq2SeqAttn(train_data)
     model.to(device)
-    train_op = optimization.get_train_op(params=model.parameters(),
-                                         hparams=config_model.opt)
+    train_op = tx.core.get_train_op(params=model.parameters(),
+                                    hparams=config_model.opt)
 
     def _train_epoch():
         data_iterator.switch_to_train_data()
+        model.train()
 
         step = 0
         for batch in data_iterator:
@@ -151,6 +146,7 @@ def main():
             data_iterator.switch_to_val_data()
         else:
             data_iterator.switch_to_test_data()
+        model.eval()
 
         refs, hypos = [], []
         for batch in data_iterator:
--- a/examples/seq2seq_attn/seq2seq_attn.py
+++ b/examples/seq2seq_attn/seq2seq_attn.py
@@ -18,10 +18,10 @@ import argparse
 import importlib
 
 import torch
+import torch.nn as nn
 
 import texar as tx
-from texar.core.optimization import get_optimizer, get_train_op
-from texar.module_base import ModuleBase
+from texar.core import optimization
 
 parser = argparse.ArgumentParser()
 parser.add_argument('--config_model',
@@ -37,8 +37,10 @@ args = parser.parse_args()
 config_model = importlib.import_module(args.config_model)
 config_data = importlib.import_module(args.config_data)
 
+device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
 
-class Seq2SeqAttn(ModuleBase):
+
+class Seq2SeqAttn(nn.Module):
 
     def __init__(self, train_data):
 
@@ -81,11 +83,14 @@ class Seq2SeqAttn(ModuleBase):
             inputs=self.source_embedder(batch['source_text_ids']),
             sequence_length=batch['source_length'])
 
+        memory = torch.cat(enc_outputs, dim=2)
+
         if mode == "train":
-            helper_train = self.decoder.create_helper()
+            helper_train = self.decoder.create_helper(
+                decoding_strategy="train_greedy")
 
             training_outputs, _, _ = self.decoder(
-                memory=torch.cat(enc_outputs, dim=2),
+                memory=memory,
                 memory_sequence_length=batch['source_length'],
                 helper=helper_train,
                 inputs=self.target_embedder(batch['target_text_ids'][:, :-1]),
@@ -101,17 +106,13 @@ class Seq2SeqAttn(ModuleBase):
             start_tokens = torch.ones_like(batch['target_length']) * \
                            self.bos_token_id
 
-            helper_infer = self.decoder.create_helper(
-                decoding_strategy="infer_greedy",
-                embedding=self.target_embedder,
+            infer_outputs = self.decoder(
                 start_tokens=start_tokens,
-                end_token=self.eos_token_id.item())
-
-            infer_outputs, _, _ = self.decoder(
-                helper=helper_infer,
-                memory=torch.cat(enc_outputs, dim=2),
+                end_token=self.eos_token_id.item(),
+                embedding=self.target_embedder,
+                memory=memory,
                 memory_sequence_length=batch['source_length'],
-                max_decoding_length=60)
+                beam_width=config_model.beam_width)
 
             return infer_outputs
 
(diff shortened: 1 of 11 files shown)


examples/seq2seq_attn/seq2seq_attn.py
```diff
@@ -101,40 +106,39 @@ def forward(self, batch, mode):
...
+    test_data = tx.data.PairedTextData(hparams=config_data.test,
+                                       device=device)
     data_iterator = tx.data.TrainTestDataIterator(
         train=train_data, val=val_data, test=test_data)
 
     model = Seq2SeqAttn(train_data)
-    optimizer = get_optimizer(model.parameters(), config_model.opt)
-    train_op = get_train_op(optimizer, config_model.opt)
+    model.to(device)
+    train_op = optimization.get_train_op(params=model.parameters(),
```
huzecong: I believe the intended usage suggested by @ZhitingHu was `tx.core.get_train_op`. Basically we only expose the first layer of packages to users.

texar/core/optimization.py
```diff
@@ -289,15 +294,22 @@ def get_train_op(optimizer: Optimizer,
...
+                         "simultaneously.")
+
+    if optimizer is None and params is not None:
+        optimizer = get_optimizer(params, hparams)
+    if optimizer is not None:
+        scheduler = get_scheduler(optimizer, hparams)
     grad_clip_fn = get_grad_clip_fn(hparams)
 
-    params_list = []
+    params_list: List = []
```
huzecong: Should the type be `List[nn.Parameter]`?

texar/modules/decoders/rnn_decoders.py
```diff
@@ -660,6 +723,55 @@ def forward(  # type: ignore
...
+
+        def _prepare_beam_search(x):
+            x = torch.unsqueeze(x, dim=1)
+            tile_dims = [1] * x.dim()
+            tile_dims[1] = beam_width
+            x = x.repeat(tile_dims)
+            shape = list(x.size())
+            shape[0] *= shape[1]
+            shape.pop(1)
+            x = x.view(shape)
```
huzecong: I think this entire function can be simplified as:
```python
x = x.unsqueeze(1).repeat(-1, beam_width, *([-1] * (x.dim() - 1)))
x = x.view(-1, *x.size()[2:])
```

texar/modules/decoders/rnn_decoders.py
```diff
@@ -660,6 +723,55 @@ def forward(  # type: ignore
...
+        def _symbols_to_logits_fn(ids, state):
+            ids = ids[:, -1]
+            inputs = embedding_fn(ids)
+            wrapper_outputs, wrapper_state = self._cell(
+                inputs, state, memory_beam_search,
+                memory_sequence_length_beam_search)
+            logits = self._output_layer(wrapper_outputs)
+            return logits, wrapper_state
+
+        outputs, log_prob = beam_search(  # type: ignore
```
huzecong: Why do we need a `# type: ignore` here?

texar/utils/beam_search.py
```diff
@@ -36,8 +36,10 @@
 INF = 1.0 * 1e7
 
 
-def gather_nd(params: torch.Tensor, indices: torch.Tensor) -> torch.Tensor:
-
+def gather_nd(params: Union[torch.Tensor, int],
```
huzecong: Why would this and following methods take `int` as an argument?

examples/seq2seq_attn/seq2seq_attn.py
ZhitingHu: Yes, let's call it by `tx.core.get_train_op` as in the Texar TF examples

examples/seq2seq_attn/seq2seq_attn.py
```diff
@@ -101,40 +106,39 @@ def forward(self, batch, mode):
             start_tokens = torch.ones_like(batch['target_length']) * \
```
ZhitingHu: Use `new_full` to create start_tokens, pls refer to PR https://github.com/asyml/texar-pytorch/pull/76

texar/utils/beam_search.py
gpengzhi: `state` used in `beam_search` is `AttentionWrapperState`. We here have `int` because of `time` in  `AttentionWrapperState`. I deleted `time` to simplify the interface in `beam_search`.

texar/modules/decoders/rnn_decoders.py
gpengzhi: `self._vocab_size` could be `None` (https://github.com/asyml/texar-pytorch/blob/master/texar/modules/decoders/decoder_base.py#L92).

texar/utils/beam_search.py
huzecong: I see. Yeah `time` is probably unnecessary, and I kind of think that some other fields in `AttentionWrapperState` are unnecessary but I'm not sure.

texar/modules/decoders/rnn_decoders.py
huzecong: In general it's much better to use assertions to guide the type checker. This adds almost no overhead anyway.

texar/utils/beam_search.py
ZhitingHu: `AttentionWrapperState` in TF [here](https://www.tensorflow.org/api_docs/python/tf/contrib/seq2seq/AttentionWrapperState) has `time`. (Is `time` used in TF?)  I'd suggest keeping it the same as the TF interface even if it's currently not used (it might be used in other attention mechanisms, probably in the future)

texar/core/optimization.py
```diff
@@ -271,12 +272,17 @@ def get_grad_clip_fn(hparams: Optional[Union[HParams,
     return grad_clip_fn
 
 
-def get_train_op(optimizer: Optimizer,
+def get_train_op(params: Optional[Union[List[torch.Tensor],
```
ZhitingHu: [Minor issue] The type of `params` looks very complex. Is it necessary or can it be simplified? (ok if not)

texar/core/optimization.py
```diff
@@ -271,12 +272,17 @@ def get_grad_clip_fn(hparams: Optional[Union[HParams,
...
+                                        List[Dict[str,
+                                                  List[torch.Tensor]]]]] = None,
+                 optimizer: Optional[Optimizer] = None,
```
ZhitingHu: `get_train_op` should also have an argument `scheduler=None` to allow users to pass in any customized scheduler, like optimizer.

texar/utils/beam_search.py
gpengzhi: If I remember correctly, in TF `time` is only used to keep track of the time step in the decoding procedure. If we really want to keep it the same as the TF interface, we could either make modifications in `beam_search` or `utils.map_structure.` (I probably can follow the similar way of `no_map_type` to bypass this issue).

examples/seq2seq_attn/seq2seq_attn.py
```diff
@@ -72,20 +73,18 @@ def __init__(self, train_data):
```
ZhitingHu:
```
tx.modules.AttentionRNNDecoder(
  ...
  input_size=self.target_embedder.dim + config_model.decoder
              ['attention']['attention_layer_size'],
```
`input_size` is not fixed yet. Any problem here?

texar/utils/beam_search.py
ZhitingHu: Yes, let's keep `time` as in TF. It's likely to be useful in the future.

examples/seq2seq_attn/seq2seq_attn.py
gpengzhi: Sorry for the late reply. I will fix #55 in this pull request as well.

ZhitingHu: Since the interface is changed, does the `AttentionRNNDecoder` doc (about `input_size`) need to be updated as well?

PR #93, opened 2019-07-07
Connector

TomNong:
1. Modify distribution connector structures, including StochasticConnector and ReparameterizedStochasticConnector;
2. Add Reparameterized distribution check in above connectors;
3. Modified utils/nest.py and utils/utils.py to fit connector functions;
4. Polish connectors code.

### DIFF
--- a/texar/modules/connectors/__init__.py
+++ b/texar/modules/connectors/__init__.py
@@ -12,13 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """
-Modules of Texar library module.
+Modules of texar library connectors.
 """
 
-from texar.modules.pretrained import *
-from texar.modules.classifiers import *
-from texar.modules.decoders import *
-from texar.modules.embedders import *  # type: ignore  # same name submodule
-from texar.modules.encoders import *
-from texar.modules.networks import *  # type: ignore  # same name submodule
-from texar.modules.connectors import *  # type: ignore  # same name submodule
+# pylint: disable=wildcard-import
+
+from texar.modules.connectors.connector_base import *
+from texar.modules.connectors.connectors import *
--- a/texar/modules/connectors/__init__.py
+++ b/texar/modules/connectors/__init__.py
@@ -15,7 +15,5 @@
 Modules of texar library connectors.
 """
 
-# pylint: disable=wildcard-import
-
 from texar.modules.connectors.connector_base import *
 from texar.modules.connectors.connectors import *
--- a/texar/modules/connectors/connector_base.py
+++ b/texar/modules/connectors/connector_base.py
@@ -15,7 +15,6 @@
 Base class for connectors that transform inputs into specified output shape.
 """
 
-# pylint: disable=arguments-differ
 from abc import ABC
 from typing import Generic, Optional, TypeVar
 from texar.hyperparams import HParams
--- a/texar/modules/connectors/connector_base.py
+++ b/texar/modules/connectors/connector_base.py
@@ -1,0 +1,74 @@
+# Copyright 2019 The Texar Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Base class for connectors that transform inputs into specified output shape.
+"""
+
+# pylint: disable=arguments-differ
+from abc import ABC
+from typing import Generic, Optional, TypeVar
+from texar.hyperparams import HParams
+from texar.module_base import ModuleBase
+
+__all__ = [
+    "ConnectorBase"
+]
+
+OutputSize = TypeVar('OutputSize')
+HParamsType = Optional[HParams]
+
+
+class ConnectorBase(ModuleBase, Generic[OutputSize], ABC):
+    r"""Base class inherited by all connector classes. A connector is to
+    transform inputs into outputs with any specified structure and shape.
+    For example, tranforming the final state of an encoder to the initial
+    state of a decoder, and performing stochastic sampling in between as
+    in Variational Autoencoders (VAEs).
+
+    Args:
+        output_size: Size of output **excluding** the batch dimension. For
+            example, set ``output_size`` to ``dim`` to generate output of
+            shape ``[batch_size, dim]``.
+            Can be an `int`, a tuple of `int`, a Tensorshape, or a tuple of
+            TensorShapes.
+            For example, to transform inputs to have decoder state size, set
+            :python:`output_size=decoder.state_size`.
+        hparams (dict, optional): Hyperparameters. Missing
+            hyperparamerter will be set to default values. See
+            :meth:`default_hparams` for the hyperparameter sturcture and
+            default values.
+    """
+
+    def __init__(self, output_size: OutputSize, hparams: HParamsType = None):
+        super().__init__(hparams)
+        self._output_size = output_size
+
+    @staticmethod
+    def default_hparams():
+        r"""Returns a dictionary of hyperparameters with default values.
+        """
+        return {
+            "name": "connector"
+        }
+
+    def forward(self, *args, **kwargs):
+        r"""Transforms inputs to outputs with specified shape.
+        """
+        raise NotImplementedError
+
+    @property
+    def output_size(self):
+        r"""The output size.
+        """
+        return self._output_size
--- a/texar/modules/connectors/connector_base.py
+++ b/texar/modules/connectors/connector_base.py
@@ -31,7 +31,7 @@ HParamsType = Optional[HParams]
 class ConnectorBase(ModuleBase, Generic[OutputSize], ABC):
     r"""Base class inherited by all connector classes. A connector is to
     transform inputs into outputs with any specified structure and shape.
-    For example, tranforming the final state of an encoder to the initial
+    For example, transforming the final state of an encoder to the initial
     state of a decoder, and performing stochastic sampling in between as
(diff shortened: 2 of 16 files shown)


texar/modules/connectors/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Modules of texar library connectors.
+"""
+
+# pylint: disable=wildcard-import
```
huzecong: This directive is redundant and can be removed. In general we don't use file-level `pylint: disable` directives unless the same warning is triggered multiple times in the file.

texar/modules/connectors/connector_base.py
```diff
@@ -0,0 +1,74 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Base class for connectors that transform inputs into specified output shape.
+"""
+
+# pylint: disable=arguments-differ
```
huzecong: Same here. Use per-line directives so it is clear where the warning is triggered.

texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,866 @@
...
+from texar.hyperparams import HParams
+from texar.modules.connectors.connector_base import ConnectorBase
+from texar.core import layers
+from texar.utils import utils
+from texar.utils import nest
+from texar.utils.types import MaybeTuple
+
+
+# pylint: disable=too-many-locals, arguments-differ
+# pylint: disable=too-many-arguments, invalid-name, no-member
```
huzecong: Same here. `too-many-*` and `invalid-name` checks are already disabled in the global configurations.

texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,866 @@
...
+        .. code-block:: python
+
+            {
+                "value": 0.,
+                "name": "constant_connector"
+            }
+
+        Here:
+
+        "value" : float
```
huzecong: The correct format for the hparams docstring is now:
```
`"value"`: float
```
The hparam name should be wrapped in backticks(`` ` ``), and there should be no spaces between the name and the colon (`:`).

texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,866 @@
...
+            Name of the connector.
+        """
+        return {
+            "value": 0.,
+            "name": "constant_connector"
+        }
+
+    def forward(self,    # type: ignore
+                batch_size: Union[int, torch.Tensor]) -> Any:
+        """Creates output tensor(s) that has the given value.
```
huzecong: All docstrings should be raw strings (`r"""..."""`).

texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,866 @@
...
+
+        output = utils.map_structure(
+            full_tensor,
+            self._output_size)
+
+        return output
+
+
+class ForwardConnector(ConnectorBase):
+    """Transforms inputs to have specified structure.
```
huzecong: Same here.

texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,866 @@
...
+
+        .. code-block:: python
+
+            # Initialized without num_samples
+            cell = LSTMCell(num_units=256)
+            # cell.state_size == LSTMStateTuple(c=256, h=256)
+            mu = torch.zeros([16, 100])
+            var = torch.ones([16, 100])
+            #var = torch.stack([torch.diag(x) for x in var], 0)
+            f_mu = torch.flatten(mu)
```
huzecong: Why do we need to flatten?

texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,866 @@
...
+                distribution_kwargs={
+                    "loc": f_mu,
+                    "scale_tril": torch.diag(f_var)})
+            output, sample = connector()
+            # output == LSTMStateTuple(c=tensor_of_shape_(1, 256),
+            #                          h=tensor_of_shape_(1, 256))
+            # sample == Tensor([16 * 100])
+
+            # Initialized with num_samples
+            connector = ReparameterizedStochasticConnector(
```
huzecong: It's unnecessary to construct a new connector.

texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,866 @@
...
+            ValueError: If distribution cannot be reparametrized.
+            ValueError: The output does not match :attr:`output_size`.
+        """
+
+        distribution = utils.check_or_get_instance(
+            self._dstr_type, self._dstr_kwargs,
+            ["torch.distributions", "torch.distributions.multivariate_normal",
+             "texar.custom"])
+
+        assert distribution.has_rsample    # type: ignore
```
huzecong: As per our discussion, please use `if ...: raise ValueError(...)` instead of `assert`.

texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,866 @@
...
+            - sample: The sample from the distribution, prior to transformation.
+
+        Raises:
+            ValueError: If distribution cannot be reparametrized.
+            ValueError: The output does not match :attr:`output_size`.
+        """
+
+        distribution = utils.check_or_get_instance(
+            self._dstr_type, self._dstr_kwargs,
+            ["torch.distributions", "torch.distributions.multivariate_normal",
```
huzecong: Do we need to include `torch.distributions.multivariate_normal`? Isn't `MultivariateNormal` included in `torch.distributions`?

texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,866 @@
...
+             "texar.custom"])
+
+        assert distribution.has_rsample    # type: ignore
+
+        if num_samples:
+            sample = distribution.rsample([num_samples])    # type: ignore
```
huzecong: Why do we need a `# type: ignore` comment? Is it possible to declare a type for `sample` that does not clash with the two assignments here?

texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,866 @@
...
+        if distribution_kwargs is None:
+            distribution_kwargs = {}
+        self._dstr_kwargs = distribution_kwargs
+
+        self._dstr = utils.check_or_get_instance(
+            distribution, self._dstr_kwargs,
+            ["torch.distributions", "torch.distributions.multivariate_normal",
+             "texar.custom"])
+
+        assert not self._dstr.has_rsample    # type: ignore
```
huzecong: Same here. Use `raise ValueError(...)`.

texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,866 @@
...
+                :attr:`num_samples` should be left as ``None``.
+            transform (bool): Whether to perform MLP transformation of the
+                distribution samples. If ``True``, the structure/shape of a
+                sample must match :attr:`output_size`. Otherwise, the size of
+                a sample will be determined by :attr:`distribution`'s
+                dimensionality.
+
+        :returns:
+            A ``Tensor`` or a (nested) ``tuple`` output, where
+            - output: A ``Tensor`` or a (nested) ``tuple`` of Tensors
```
huzecong: There should probably be a new line before a list.

texar/modules/connectors/connectors_test.py
```diff
@@ -0,0 +1,212 @@
+#
```
huzecong: This line is unnecessary.

texar/utils/nest.py
```diff
@@ -0,0 +1,240 @@
+# Apply from
+# pylint:disable=line-too-long
+# Tensorflow(https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/util/nest.py)
```
huzecong: URLs, if on their own line, will not trigger the `line-too-long` warning.

texar/utils/utils.py
```diff
@@ -125,6 +125,8 @@ def map_structure(fn: Callable[[T], R], obj: Collection[T]) -> Collection[R]:
     if isinstance(obj, list):
         return [map_structure(fn, x) for x in obj]
     if isinstance(obj, tuple):
+        if isinstance(obj, torch.Size):
+            return fn(obj)
```
huzecong: Could you also change `map_structure_zip` accordingly?

texar/modules/connectors/connectors.py
TomNong: Yeah... There will be error "error: `T" has no attribute "rsample"` if not `# type: ignore`.

texar/modules/connectors/connectors.py
TomNong: This part is corrected, thanks!

texar/modules/connectors/connectors.py
huzecong: I see. I think the problem here is that we should not use a generic type `T` but rather just the `torch.Distribution` type. Could you try to see if this gets rid of the error?

huzecong:
Please mark comments as resolved after having addressed them in commits.

texar/modules/connectors/connectors.py
```diff
@@ -0,0 +1,866 @@
...
+
+        if num_samples:
+            sample = dstr.rsample([num_samples])
+        else:
+            sample = dstr.rsample()
+
+        if transform:
+            fn_modules = ['torch', 'torch.nn', 'texar.custom']
+            activation_fn = utils.get_function(
+                self.hparams.activation_fn, fn_modules)
```
huzecong: It's better to move this part into `__init__`. Also, we should probably include `torch.nn.functional` instead of `torch.nn` because we're using the functional APIs.

huzecong: And yeah, we have `texar.core.get_activation_fn`, why not just use that? Could you also change `get_activation_fn` to include `torch` in `fn_modules`

TomNong: Updated, thanks!

PR #99, opened 2019-07-11
Add GPT-2 modules & Polish GPT-2 examples 

gpengzhi:
1. add `GPT2Base`, `GPT2Encoder`, `GPT2Decoder`, and `GPT2Classifier`. The code structures of `GPT2Base`, `GPT2Encoder`, and `GPT2Classifier` follow a similar way as that in `Bert` Modules. `GPT2Decoder` has the same interface as that in `TransformerDecoder`. 

2. Polish the `gpt-2` example so that it is implemented by the lastest `GPT2` modules.


examples/gpt-2/configs/config_model_117M.py
```diff
@@ -1,30 +1,30 @@
 """Texar config file of the GPT-2 model_117M model.
 """
+pretrained_model_name = "117M"
 
 vocab_size = 50257
-dim = 768
```
ZhitingHu: I think it's better to define `dim=768` and use it (instead of `768`) in the following? This makes it clear all places using `dim` must be of the same value.

ZhitingHu: Another question, is it even necessary to have these model config files at all? Our built-in `GPT2Decoder` should already have defined the hyperparameters, right (user only has to specify 117M or 345M)?

examples/gpt-2/README.md
```diff
@@ -1,31 +1,18 @@
 # GPT-2: Pre-trained Langauge Model
```
ZhitingHu: Could you update the code examples in the `Other Use Cases` section of README to fit the lastest interfaces?

examples/gpt-2/configs/config_model_117M.py
gpengzhi: The reason that I have the model config file is that we allow users to load their own checkpoint. They could write the model architecture in the model config file and set `pretrained_model_name` to be `None`. They then can use `torch.load()` to load their own checkpoint to their own defined model.

gpengzhi: The reason I deleted `dim=768` is that I use this model config file as `hparams` for `GPT2Decoder` directly. In the default `hparams` of `GPT2Decoder`, there is no corresponding field called `dim`.

ZhitingHu: So it looks *loading (pretrained) model checkpoint* and *model arch config* are coupled? I.e., setting `pretrained_model_name` will not only define the model arch but also load the pretrained checkpoint. Can we decouple them, by having another hparam control whether to load pretrained checkpoint or not?

ZhitingHu:
Pls also add an option to `gpt2_generate_main.py` to allow users to use top-p decoding (with `TopPSampleEmbeddingHelper`). Currently the code only allows top-k decoding (with `TopKSampleEmbeddingHelper`)

ZhitingHu:
Pls update the README too. I believe the model download section is no longer needed?

examples/gpt-2/gpt2_generate_main.py
```diff
@@ -135,42 +78,29 @@ def run_model():
...
-        gpt2_config = importlib.import_module(
-            args.config_model)
-    else:
-        raise ValueError('Unknown config_type.')
-
-    assert max_decoding_length <= gpt2_config.position_size, (
+    assert max_decoding_length <= config_model["position_size"], (
         "max_decoding_length should not be greater than position size")
     assert nsamples % batch_size == 0, (
         "nsamples must be dividable by batch_size")
```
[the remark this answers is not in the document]
huzecong: Also, this assertion here does not make sense -- it is unnecessary for `nsamples` to be divisible (not "dividable") by `batch_size`. I'm guessing this was there because of computation graph issues with TF.

PR #100, opened 2019-07-12
Add main method for xlnet text generation

AvinashBukkittu:
This PR adds
  - An XLNet text generation main method from which user can play with XLNet either interactively or non-interactively.

### DIFF
--- a/examples/xlnet/xlnet_generation_main.py
+++ b/examples/xlnet/xlnet_generation_main.py
@@ -1,0 +1,145 @@
+# Copyright 2019 The Texar Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Example of building XLNet language model for sample generation.
+"""
+
+import argparse
+import torch
+import sentencepiece as spm
+
+import xlnet
+import xlnet.model.decoder
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--checkpoint', type=str, default=None,
+                    help="Checkpoint to load model weights from. Use "
+                         "`--pretrain_checkpoint` instead if loading XLNet "
+                         "pretrained checkpoint.")
+parser.add_argument('--pretrain_checkpoint', type=str,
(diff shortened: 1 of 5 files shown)


examples/xlnet/xlnet_generation_main.py
```diff
@@ -0,0 +1,145 @@
...
+        decode_samples = decode_output.sample_id.tolist()
+        for idx, sample_tokens in enumerate(decode_samples):
+            print(f"=== Sample {idx} ===")
+            output = "\n".join(sp_model.DecodeIds(xs) for xs in split_by(
+                sample_tokens, xlnet.data.utils.special_symbols["<eop>"]))
+            print(output)
+
+    if args.is_interactive:
+        try:
+            from IPython import embed
```
ZhitingHu: There shouldn't be `IPython`. Please make everything the same as [gpt2_generate_main.py](https://github.com/asyml/texar-pytorch/blob/8ce8589120dc54bcd28c8f9458128422f998bdf2/examples/gpt-2/gpt2_generate_main.py), except that here we use XLNet instead of GPT-2.

ZhitingHu: E.g., In gpt2-generate-main, https://github.com/asyml/texar-pytorch/blob/8ce8589120dc54bcd28c8f9458128422f998bdf2/examples/gpt-2/gpt2_generate_main.py#L109-L112

Users only type in the prompt text (instead of the code `sample(...)`)

AvinashBukkittu: GPT2 has interactive option. Should this file also have interactive option but without using IPython?

AvinashBukkittu: Got it.

ZhitingHu: yep. we could just copy-paste the interactive code from gpt2

AvinashBukkittu: Removed IPython dependency in f3ceca0

ZhitingHu:
Thanks. Now, `xlnet_generate_main.py` allows only top-k decoding (with `TopKSampleEmbeddingHelper`). Please also add an option to allow use of top-p decoding (with `TopPSampleEmbeddingHelper`). I left the same comment to the PR https://github.com/asyml/texar-pytorch/pull/99. Pls work with @gpengzhi, and add the option to gpt2 and xlnet, respectively

examples/xlnet/xlnet_generation_main.py
```diff
@@ -0,0 +1,157 @@
...
+"""
+
+import argparse
+import torch
+import sentencepiece as spm
+
+import xlnet
+import xlnet.model.decoder
+from texar.modules.decoders.decoder_helpers import TopKSampleEmbeddingHelper, \
+    TopPSampleEmbeddingHelper
```
huzecong: These should be imported from `texar.modules`. The previous discussion was to not expose too much internal details to the user.

AvinashBukkittu: Done in 0e23eda

PR #101, opened 2019-07-12
Add output size to Texar modules

TomNong:
Add property `output_size` to `Modulebase` and its sub-modules;
Modify sub-modules in `modules` to adjust the computation.

### DIFF
--- a/texar/modules/classifiers/conv_classifiers.py
+++ b/texar/modules/classifiers/conv_classifiers.py
@@ -241,3 +241,20 @@ class Conv1DClassifier(ClassifierBase):
         r"""A list of uniquified layer names.
         """
         return self._encoder.layer_names
+
+    @property
+    def output_size(self) -> int:
+        r"""The final dimension(s) of :meth:`forward` output tensor(s).
+
+        Here output is :attr:`logits`. The final dimension equals to ``1``
+        when output final dimension is only determined by input.
+        """
+        if self._hparams.num_classes > 1:
+            logit_dim = self._hparams.num_classes
+        elif self._hparams.num_classes == 1:
+            logit_dim = 1
+        else:
+            raise AttributeError("'Conv1DClassifier' object has"
+                                 "no attribute 'output_size'"
+                                 "if 'self._hparams.num_classes' < 1.")
+        return logit_dim
--- a/texar/modules/embedders/embedder_base.py
+++ b/texar/modules/embedders/embedder_base.py
@@ -142,3 +142,12 @@ class EmbeddingDropout(ModuleBase):
         mask += input_tensor.new_empty(noise_shape).uniform_(0, 1)
         mask = torch.floor(mask).div_(keep_rate)
         return input_tensor * mask
+
+    @property
+    def output_size(self) -> int:
+        r"""The final dimension(s) of :meth:`forward` output tensor(s).
+
+        Here final dimension is ``1``, because here output tensor size equals
+        to the input tensor size.
+        """
+        return 1
--- a/texar/modules/encoders/multihead_attention.py
+++ b/texar/modules/encoders/multihead_attention.py
@@ -255,6 +255,4 @@ class MultiheadAttentionEncoder(EncoderBase):
 
     @property
     def output_size(self):
-        r"""Provides output dimension as property.
-        """
         return self._hparams.output_dim
--- a/texar/modules/encoders/rnn_encoders.py
+++ b/texar/modules/encoders/rnn_encoders.py
@@ -460,6 +460,18 @@ class UnidirectionalRNNEncoder(RNNEncoderBase[State]):
         """
         return self._output_layer
 
+    @property
+    def output_size(self) -> int:
+        # TODO: We will change the implementation to
+        # something that does not require a forward pass.
+
+        dim = self._cell.hidden_size
+        dummy_tensor = torch.Tensor(dim)
+        if self._output_layer is not None:
+            return self._output_layer(dummy_tensor).size(-1)
+        else:
+            return dim
+
 
 class BidirectionalRNNEncoder(RNNEncoderBase):
     r"""Bidirectional forward-backward RNN encoder.
@@ -788,3 +800,26 @@ class BidirectionalRNNEncoder(RNNEncoderBase):
         r"""The output layer of the backward RNN.
         """
         return self._output_layer_bw
+
+    @property
+    def output_size(self) -> Tuple[int, int]:
+        r"""The final dimension(s) of :meth:`forward` output tensor(s).
+
+        Here final dimension is a tuple consisting of forward
+        RNN final dimension and backward RNN final dimension.
+        """
+        # TODO: We will change the implementation to
+        # something that does not require a forward pass.
+        dim_bw = self._cell_bw.hidden_size
+        dummy_tensor_bw = torch.Tensor(dim_bw)
+        dim_fw = self._cell_fw.hidden_size
+        dummy_tensor_fw = torch.Tensor(dim_fw)
+        if self._output_layer_bw is not None:
+            output_bw = self._output_layer_bw(dummy_tensor_bw).size()[-1]
+        else:
+            output_bw = dim_bw
+        if self._output_layer_fw is not None:
+            output_fw = self._output_layer_fw(dummy_tensor_fw).size()[-1]
+        else:
+            output_fw = dim_fw
+        return (output_fw, output_bw)
--- a/texar/modules/networks/conv_networks.py
+++ b/texar/modules/networks/conv_networks.py
@@ -531,3 +531,14 @@ class Conv1DNetwork(FeedForwardNetworkBase):
         with torch.no_grad():
             output = super().forward(input)
         return output.view(output.size()[0], -1).size()
+
+    @property
+    def output_size(self) -> int:
+        if self.hparams.num_dense_layers <= 0:
+            return self._hparams.out_channels * len(self._hparams.kernel_size)
+        else:
+            out_features = self._hparams.out_features
+            if isinstance(out_features, (list, tuple)):
+                return out_features[-1]
+            else:
+                return out_features
--- a/texar/modules/networks/networks.py
+++ b/texar/modules/networks/networks.py
@@ -16,6 +16,7 @@ Various neural networks and related utilities.
 """
 
 from texar.modules.networks.network_base import FeedForwardNetworkBase
+from texar.utils.utils import get_output_size
 
 __all__ = [
     "FeedForwardNetwork",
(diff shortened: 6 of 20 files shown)


texar/modules/classifiers/classifier_base.py
```diff
@@ -40,3 +40,7 @@ def default_hparams() -> Dict[str, Any]:
 
     def forward(self, *input: torch.Tensor):  # pylint: disable=redefined-builtin
         raise NotImplementedError
+
+    @property
+    def output_size(self) -> int:
```
huzecong: I don't think it's necessary to add this implementation as long as the class is an `ABC`. Does it fail the CI otherwise?

texar/modules/embedders/embedder_base.py
```diff
@@ -101,6 +101,12 @@ def num_embeds(self) -> int:
         """
         return self._num_embeds
 
+    @property
+    def output_size(self) -> int:
+        r"""Provides output dimension as property.
```
huzecong: Please be consistent in wording.

texar/modules/encoders/rnn_encoders.py
```diff
@@ -460,6 +461,17 @@ def output_layer(self) -> Optional[nn.Module]:
...
         return self._output_layer
 
+    @property
+    def output_size(self) -> int:
+        r"""The output size.
+        """
+        dim = self._cell.hidden_size
+        dummy_tensor = torch.Tensor(dim)
+        if self._output_layer is not None:
+            return self._output_layer(dummy_tensor).size()[-1]
```
huzecong: Use `.size(-1)`. And please add a TODO here saying we'll change the implementation to something that does not require a forward pass.

texar/modules/encoders/rnn_encoders.py
```diff
@@ -788,3 +801,21 @@ def output_layer_bw(self) -> Optional[nn.Module]:
         r"""The output layer of the backward RNN.
         """
         return self._output_layer_bw
+
+    @property
+    def output_size(self) -> tuple:
```
huzecong: The type should be `Tuple[int, int]`.

texar/modules/networks/networks.py
```diff
@@ -82,3 +82,9 @@ def default_hparams():
             "layers": [],
             "name": "NN"
         }
+
+    @property
+    def output_size(self) -> int:
+        r"""The output size.
+        """
+        return self._layers[-1].out_features
```
huzecong: What if the final layer is not `nn.Linear`?

TomNong: Yeah actually the docs says `self._layers` is a list of `nn.Linear` here...

texar/modules/encoders/multihead_attention.py
```diff
@@ -255,6 +255,6 @@ def _combine_heads(self, x: torch.Tensor) -> torch.Tensor:
 
     @property
     def output_size(self):
-        r"""Provides output dimension as property.
+        r"""Return the finale dimension of :meth:`forward` output tensor.
```
huzecong: Should be "final" but not "finale". "Finale" means the last piece of music in a performance.

texar/modules/encoders/rnn_encoders.py
```diff
@@ -788,3 +804,23 @@ def output_layer_bw(self) -> Optional[nn.Module]:
         r"""The output layer of the backward RNN.
         """
         return self._output_layer_bw
+
+    @property
+    def output_size(self) -> Tuple[int, int]:
+        r"""Return the finale dimensions of :meth:`forward` output tensor.
```
huzecong: Could we unify the docstrings for all modules? As long as the base method `ModuleBase.output_size` has docstrings, I think it's okay to omit docstrings for certain modules if they don't require extra explanantions.

texar/modules/networks/conv_networks.py
```diff
@@ -531,3 +533,19 @@ def _infer_dense_layer_input_size(self, input: torch.Tensor) -> torch.Size:
...
+    @property
+    def output_size(self) -> int:
+        r"""Return the finale dimension of :meth:`forward` output tensor.
+        """
+        # TODO: We will change the implementation to
+        # something that does not require a forward pass.
+        if self._in_features is not None:
+            input = torch.Tensor(1, self._in_channels, self._in_features)
+        else:
+            kernel_size = self._hparams.kernel_size
```
huzecong: Do we have to do a forward pass here? `input_size` to the dense layers are computed in `__init__` via a forward pass, can use just reuse that result?

texar/modules/networks/network_base.py
```diff
@@ -134,6 +134,12 @@ def layer_names(self) -> List[str]:
         """
         return self._layer_names
 
+    @property
+    def output_size(self) -> int:
+        r"""Return the finale dimension of :meth:`forward` output tensor.
```
huzecong: Again, I don't think it's necessary to implement an abstract method as an abstract method. Also, this module is not marked as ABC, which means it is a concrete class. Do `output_size` computations differ for all its derived classes?

texar/utils/utils.py
```diff
@@ -103,7 +106,7 @@ def no_map(container_type: Type[T], *args, **kwargs) -> T:
         args: Arguments to the constructor.
         kwargs: Keyword arguments to the constructor
 
-    Returns:
+    :returns:
```
huzecong: There's no need to change `Returns:` to `:returns:` if the return value description does not contain colons (`:`).

texar/utils/utils.py
```diff
@@ -605,6 +608,37 @@ def get_instance_kwargs(kwargs: Kwargs, hparams: ParamDict) -> Kwargs:
...
+
+    If type of :attr:`instance_` is among the common types, the final
+    dimension of output size will be computed.
+
+    Args:
+        instance_: A :class:`~torch.nn.Module` instance from
+            which to compute output size.
+
+    :returns:
+        int (optional): The final dimension of the output size.
```
huzecong: And here, what's in front of the colon (`:`) is exactly the type description, which is the legitimate use of colons in a `Returns:` directive.

texar/utils/utils.py
```diff
@@ -605,6 +608,37 @@ def get_instance_kwargs(kwargs: Kwargs, hparams: ParamDict) -> Kwargs:
     return kwargs_
 
 
+def get_output_size(instance_: Optional[Module]) -> Optional[int]:
```
huzecong: It's strange to accept a `None` as argument. Are there any use cases where this would be more convenient?

texar/utils/utils.py
```diff
@@ -605,6 +608,37 @@ def get_instance_kwargs(kwargs: Kwargs, hparams: ParamDict) -> Kwargs:
...
+        nn.Conv2d: lambda x: x.out_channels * len(x.kernel_size),
+        nn.Conv3d: lambda x: x.out_channels * len(x.kernel_size),
+        nn.Embedding: lambda x: x.embedding_dim,
+        nn.EmbeddingBag: lambda x: x.embedding_dim,
+        nn.RNNCell: lambda x: x.hidden_size,
+        nn.LSTMCell: lambda x: x.hidden_size,
+        nn.GRUCell: lambda x: x.hidden_size,
+    }
+    for t, l in types.items():
+        if isinstance(instance_, t):
```
huzecong: Consider replacing `Conv*d` with the more general `_ConvNd` base class. Since `_ConvNd` is also the base class for `ConvTranspose*d` classes, please also take a look at their code to make sure the output size calculation is correct. If their sizes are not computed in the same way, let's keep this part as is.

The `*Cell` classes can be with `RNNCellBase` without errors.

texar/utils/utils.py
```diff
@@ -605,6 +608,37 @@ def get_instance_kwargs(kwargs: Kwargs, hparams: ParamDict) -> Kwargs:
...
+    dimension of output size will be computed.
+
+    Args:
+        instance_: A :class:`~torch.nn.Module` instance from
+            which to compute output size.
+
+    :returns:
+        int (optional): The final dimension of the output size.
+    """
+    types = {
```
huzecong: Let's move this dictionary outside the method to prevent it from being constructed every time the method is called.

texar/modules/embedders/embedder_base.py
```diff
@@ -142,3 +148,12 @@ def forward(self,  # type: ignore
         mask += input_tensor.new_empty(noise_shape).uniform_(0, 1)
         mask = torch.floor(mask).div_(keep_rate)
         return input_tensor * mask
+
+    @property
+    def output_size(self) -> int:
+        r"""The output size, here output size equals to the input tensor size.
+
+        :returns:
+            - 1, since the output dimensions are equals to input dimensions.
```
huzecong: No need to use a list when we have only one bullet point. This is also confusing because it could be read as `-1`.

texar/modules/classifiers/conv_classifiers.py
```diff
@@ -241,3 +241,14 @@ def layer_names(self) -> List[str]:
         r"""A list of uniquified layer names.
         """
         return self._encoder.layer_names
+
+    @property
+    def output_size(self) -> int:
+        r"""Return the final dimension of the output :attr:`logit`,
+        if the final dimension cannot be computed, returns 1.
```
huzecong: Do we return `-1` here? Now if `self._hparams.num_classes < 1` we would just run into a `NameError` because `logit_dim` would not be defined.

texar/modules/classifiers/conv_classifiers.py
```diff
@@ -241,3 +241,17 @@ def layer_names(self) -> List[str]:
         r"""A list of uniquified layer names.
         """
         return self._encoder.layer_names
+
+    @property
+    def output_size(self) -> int:
+        r"""Return the final dimension of the output :attr:`logit`,
+        if the final dimension cannot be computed, return 1.
```
ZhitingHu: The doc is inconsistent with the actual implementation when "the final dimension cannot be computed"

ZhitingHu: Also, for a class property (`@property`), the docstring usually does not say "Return xxx"

texar/modules/classifiers/conv_classifiers.py
```diff
@@ -241,3 +241,17 @@ def layer_names(self) -> List[str]:
...
+        """
+        if self._hparams.num_classes > 1:
+            logit_dim = self._hparams.num_classes
+        elif self._hparams.num_classes == 1:
+            logit_dim = 1
+        else:
+            raise NameError("logit_dim cannot be defined "
+                            "if self._hparams.num_classes < 1")
```
ZhitingHu: 1. The error msg should not refer to `logit_dim` which is a local variable and should not be exposed to users.

2. I don't think a `NameError` is appropriate here. Probably `AttributeError`

texar/utils/utils.py
huzecong: This is not resolved. The `:returns:` here could be changed back to `Returns:` without problem. Could you check to see if this is the case?

TomNong: Oh sorry, I thought the content containing the colon (:) needed to labeled `:returns:`, I'll fix this, thanks!

huzecong:
Please, only force-push when necessary. We were going to squash commits anyway so there's no need to force-push for every little change. It's almost impossible for me to know what you changed without going through the entire code diff.

Issue #90, opened 2019-07-02
Integrate bleu_tool.py into texar.evals


ZhitingHu:
related issue https://github.com/asyml/texar-pytorch/issues/88

gpengzhi:
Reopen this issue to resolve the comments in https://github.com/asyml/texar-pytorch/pull/105.

PR #104, opened 2019-07-15
Use texar.evals directly in Transformer example

gpengzhi:
#90

### DIFF
--- a/texar/evals/bleu.py
+++ b/texar/evals/bleu.py
@@ -206,10 +206,10 @@ def corpus_bleu(list_of_references: List[List[MaybeList[str]]],
         if ratio > 1.0:
             bp = 1.
         else:
-            try:
+            if abs(ratio) < 1e-8:
+                bp = 0.
+            else:
                 bp = math.exp(1 - 1. / ratio)
-            except ZeroDivisionError:
-                bp = math.exp(1 - 1. / (ratio + 1e-8))
     else:
         bp = 1.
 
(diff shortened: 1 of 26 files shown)


texar/evals/bleu.py
```diff
@@ -196,15 +201,17 @@ def corpus_bleu(list_of_references: List[List[MaybeList[str]]],
...
-    if ratio > 1.0:
-        bp = 1.
+    if use_bp:
+        ratio = float(hypothesis_length) / reference_length
+        if ratio > 1.0:
+            bp = 1.
+        else:
+            try:
+                bp = math.exp(1 - 1. / ratio)
+            except ZeroDivisionError:
```
huzecong: This try-catch is really inelegant (and potentially a tiny bit slower). Let's just do a check on `abs(ratio) < 1e-8`.

gpengzhi:
@huzecong Yes. I tested them on some simple examples. The only difference is that `BLEU` score is multiplied by 100 in `texar` implementation.

ZhitingHu:
I believe the original `blue_tool` usually gets higher BLEU than the moses bleu (tx.evals.corpus_bleu_moses). So `blue_tool = tx.evals.corpus_bleu > tx.evals.corpus_bleu_moses`?
Can you double check?

ZhitingHu:
If it's true, let's add this to the doc of `tx.evals.corpus_bleu` and `sentence_bleu`, sth like

"""
This BLEU has been used in evaluating Transformer (Vaswani et al) "Attention is all you need" for machine translation. The resulting BLEU numbers are usually a bit higher than `texar.evals.corpus_bleu_moses`
"""

PR #107, opened 2019-07-19
Add XLNet Modules

### LINKED ISSUE #106: Add configuration for XLNet-Base
The XLNet authors have just released a new pre-trained model checkpoint named `XLNet-Base, Cased` (https://github.com/zihangdai/xlnet#released-models). We should add support for this checkpoint.

gpengzhi:
1. This PR includes `XLNetBase`, `XLNetEncoder`, `XLNetDecoder`, `XLNetClassifier`, and `XLNetRegressor`.
2. Resolve #106


gpengzhi:
@AvinashBukkittu For your reference.

texar/modules/encoders/xlnet_encoder.py
```diff
@@ -0,0 +1,480 @@
...
+        self.r_w_bias: Optional[nn.Parameter]
+        self.r_s_bias: Optional[nn.Parameter]
+
+        if not self._hparams.untie_r:
+            self.r_r_bias = nn.Parameter(torch.Tensor(num_heads, head_dim))
+            self.r_w_bias = nn.Parameter(torch.Tensor(num_heads, head_dim))
+            self.r_s_bias = (nn.Parameter(torch.Tensor(num_heads, head_dim))
+                             if self._hparams.use_segments else None)
+        else:
+            self.r_r_bias = None
```
huzecong: Something I just learnt last week: if some variable should have type `Optional[T]`, instead of writing it like this, we can just do:
```python
var = None
if ...:
    var = something_else
```
This is much more concise and can also pass type-check.

texar/modules/pretrained/__init__.py
```diff
@@ -19,3 +19,6 @@
 from texar.modules.pretrained.bert_utils import *
 from texar.modules.pretrained.gpt2_base import *
 from texar.modules.pretrained.gpt2_utils import *
+from texar.modules.pretrained.xlnet_base import *
+from texar.modules.pretrained.xlnet_model_utils import *
```
huzecong: Do we need to export `xlnet_model_utils`? I think these are modules used only in XLNet.

texar/modules/pretrained/xlnet_utils.py
```diff
@@ -0,0 +1,247 @@
...
+        print(f"WARNING: Certain weights from checkpoint are not loaded: "
+              f"{list(from_params.keys())}")
+
+    filtered_to_params = [k for k in to_params if k.startswith("xlnet")]
+    if len(filtered_to_params) > 0:
+        print(f"WARNING: Certain parameters are not initialized: "
+              f"{list(filtered_to_params)}")
+
+
+def _default_download_dir() -> str:
```
huzecong: This same method is written in every `*_utils` file. Please extract them into another file (e.g. `pretrained_utils.py`) and reuse the code.

texar/modules/pretrained/xlnet_utils.py
```diff
@@ -0,0 +1,247 @@
...
+        texar_download_dir = os.path.join(home_dir, 'texar_download')
+
+    if not os.path.exists(texar_download_dir):
+        os.mkdir(texar_download_dir)
+
+    return os.path.join(texar_download_dir, 'xlnet')
+
+
+def load_pretrained_model(pretrained_model_name: str,
+                          cache_dir: Optional[str] = None) -> str:
```
huzecong: And this method also. Let's pass the filename and/or cache path as arguments.

texar/modules/regressors/xlnet_regressor.py
```diff
@@ -0,0 +1,244 @@
...
+from texar.modules.encoders.xlnet_encoder import XLNetEncoder
+from texar.utils import utils
+
+
+__all__ = [
+    "XLNetRegressor",
+]
+
+
+class XLNetRegressor(RegressorBase):
```
huzecong: Do we really need a separate regressor here? I implemented a regressor because both it and the classifier could share a common base class `XLNetSummary`. Now, with this middle layer is gone, there's a lot of duplicated code among the two classes.
Is it possible to do regression using classifiers in Texar (e.g. by setting `num_classes` to zero)? If it is, I think we can remove this class.

texar/modules/classifiers/xlnet_classifier.py
```diff
@@ -0,0 +1,287 @@
...
+                    time steps.
+
+            `use_projection`: bool
+                If `True`, an additional `Linear` layer is added after the
+                summary step.
+
+            `num_classes`: int
+                Number of classes:
+
+                - If **> 0**, an additional `Linear`
```
huzecong: Let's use `` :torch_nn:`Linear` ``.

texar/modules/classifiers/xlnet_classifier.py
```diff
@@ -0,0 +1,287 @@
...
+        Here:
+
+        1. Same hyperparameters as in
+            :class:`~texar.modules.XLNetEncoder`.
+            See the :meth:`~texar.modules.XLNetEncoder.default_hparams`.
+            An instance of XLNetEncoder is created for feature extraction.
+
+        2. Additional hyperparameters:
+
+            `summary_type`: str
```
huzecong: The hyperparameter should be formatted as `` `"name"`: type ``. Note the quotes inside the backticks. Although this does not affect the docstring syntax, we keep things this way for consistency.

texar/modules/regressors/xlnet_regressor.py
gpengzhi: I think `XLNetRegressor` is necessary here. I added some new logic to `XLNetClassifier` following the logic of `BertClassifier`. If `num_classes <= 0` in `XLNetClassifer`, we do not append the dense layer, and the number of classes is assumed to be the final dense layer size of the encoder.

gpengzhi:
Please ignore my last commit.

texar/modules/regressors/xlnet_regressor.py
gpengzhi: It is true that the code structure of `XLNetRegressor` and `XLNetClassifier` are very similar. But this might be a cleaner way to present our modules. Please let me know your thought.

texar/modules/encoders/xlnet_encoder.py
```diff
@@ -0,0 +1,480 @@
...
+        return ret
+
+    def forward(self,  # type: ignore
+                token_ids: torch.LongTensor,
+                *args,
+                **kwargs) \
+            -> Tuple[torch.Tensor, Optional[List[torch.Tensor]]]:
+        r"""A wrapper for :meth:`_forward`. This layer exists because
+        :class:`XLNetDecoder` compute embeddings in the decoder helper.
+        Please refer to :meth:`_forward` for the full list of arguments.
```
ZhitingHu: Does this mean `XLNetDecoder` shares `_forward` w/ `XLNetEncoder`?

If so, this is kind of internal implementation details and perhaps should not be exposed to users. I'd suggest moving the docstring of `_forward` to `forward`. And `_forward` is not visible to users.

texar/modules/pretrained/xlnet_model_utils.py
```diff
@@ -0,0 +1,342 @@
...
+from texar.module_base import ModuleBase
+
+__all__ = [
+    "PositionWiseFF",
+    "RelativeMultiheadAttention",
+    "RelativePositionalEncoding",
+]
+
+
+class PositionWiseFF(ModuleBase):
```
ZhitingHu: Is it the same as the position_wise networks used in transformer models? https://github.com/gpengzhi/texar-pytorch/blob/xlnet/texar/modules/encoders/transformer_encoder.py#L36

If so, let's use the same way to construct this part. I'd prefer the previous way that uses texar FeedForwardNet

texar/modules/pretrained/xlnet_model_utils.py
```diff
@@ -0,0 +1,342 @@
...
+        output = self.activation_fn(output)
+        output = self.dropout(output)
+        output = self.linear2(output)
+        output = self.dropout(output)
+        # residual + layer norm
+        output = self.layer_norm(input + output)
+        return output
+
+
+class PositionalEmbedding(nn.Module):
```
ZhitingHu: `texar.modules.embeddings` already has Position Embedders. If this is a different embedder, pls make a new Embedder class in `texar.modules.embeddings`, following the conventions of Position Embedders as much as possible

texar/modules/pretrained/xlnet_model_utils.py
```diff
@@ -0,0 +1,342 @@
...
+        self.register_buffer('inv_freq', inv_freq)
+
+    def forward(self,  # type: ignore
+                pos_seq: torch.Tensor) -> torch.Tensor:
+        sinusoid = torch.ger(pos_seq, self.inv_freq)
+        pos_embed = torch.cat([sinusoid.sin(), sinusoid.cos()], dim=-1)
+        return pos_embed
+
+
+class RelativePositionalEncoding(ModuleBase):
```
ZhitingHu: Same for this position embedder.

It's possible that these embedders are usable in other (future) models. In the docstring, we can say "this embedder is used in XLNet"

texar/modules/pretrained/xlnet_model_utils.py
```diff
@@ -0,0 +1,342 @@
...
+            bwd_pos_embed = self._create_positional_embedding(
+                -start, -end, 1, batch_size // 2, clamp_len)
+            pos_embed = torch.cat([fwd_pos_embed, bwd_pos_embed], dim=1)
+        else:
+            pos_embed = self._create_positional_embedding(
+                start, end, -1, batch_size, clamp_len)
+        return pos_embed
+
+
+class RelativeMultiheadAttention(ModuleBase):
```
ZhitingHu: Is this module possibly applicable in other models? What's the difference from `MultiheadAttentionEncoder`? We can discuss this in person.

It's one of the key advantages of Texar over other libraries that we correctly factorize out any re-usable components and make them as standardized modules (e.g., the above new embedders)

texar/modules/pretrained/xlnet_utils.py
```diff
@@ -0,0 +1,247 @@
...
+    configs["activation"] = config_ckpt["ff_activation"]
+    configs["num_heads"] = config_ckpt["n_head"]
+    configs["num_layers"] = config_ckpt["n_layer"]
+    configs["vocab_size"] = config_ckpt["n_token"]
+    configs["untie_r"] = config_ckpt["untie_r"]
+
+    return configs
+
+
+def sum_tensors(xs: List[Optional[torch.Tensor]]) -> Optional[torch.Tensor]:
```
ZhitingHu: I think this can go to `texar.utils`

texar/modules/regressors/xlnet_regressor.py
ZhitingHu: I'd suggest keeping regressor modules separate from classifiers. But we shall reduce the code duplication, e.g., by sharing the code portion in `__init__` (encoder-to-logits) possibly by wrapping it as a function.

texar/modules/pretrained/xlnet_base.py
```diff
@@ -0,0 +1,96 @@
...
+from texar.module_base import ModuleBase
+from texar.modules.pretrained import xlnet_utils
+
+
+__all__ = [
+    "XLNetBase",
+]
+
+
+class XLNetBase(ModuleBase):
```
AvinashBukkittu: Just a thought - Why do we need XLNetBase and BertBase classes? Can we not have XLNetEncoder and BERTEncoder inherit from EncoderBase and avoid this inheritance? It seems to this, this hierarchy may be a bit unnecessary.

texar/modules/classifiers/xlnet_classifier.py
```diff
@@ -0,0 +1,287 @@
...
+                The summary type, one of:
```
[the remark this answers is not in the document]
huzecong: I think we could use the same. By default XLNet uses `last` summary type, which is `cls_time` in Texar (but uses state from final time step because XLNet is trained with `CLS` token at the end). Other strategies (`mean`, `attn`) were implemented in their repo but were not mentioned or used, so they probably don't perform well.

texar/modules/encoders/xlnet_encoder.py
huzecong: Yes -- the reason to have a separate `_forward` method was because `Helper` returns the embeddings instead of tokens. I agree we should not expose internal methods and we should move the docstring.

texar/modules/pretrained/xlnet_model_utils.py
huzecong: This is almost the exact positional embedder as the one we have in Texar, but they're slightly different. Texar uses `num_timescales - 1` as denominator in the exponent while here it is just `num_timescales`. I'm wondering what was the intention of using `num_timescales - 1`, because I think (and correct me if I'm wrong) the implementation here is the orthodox version.

texar/modules/pretrained/xlnet_model_utils.py
huzecong: TBH this module is specific to XLNet, and all it does is create sinusoidal positional embeddings for specific index ranges. I think we should keep this only if we add `RelativeMultiheadAttention` to Texar modules.

texar/modules/regressors/xlnet_regressor.py
huzecong: Sharing `__init__` would require `XLNetRegressor` to inherit from `XLNetClassifier`, which I think is kind of weird. Merging the regressor interface with classifiers would mean that all existing classifiers (Conv, BERT, GPT2) could be used as regressors out-of-the-box.

texar/modules/pretrained/xlnet_base.py
huzecong: I think the point of having `XLNetBase` (and `BertBase` similarly) is to separate the logic for loading pretrained models. That said, I feel the loading logic is the same for almost all pretrained models, so maybe we can unify them as something like `PretrainedBase`?

texar/modules/pretrained/xlnet_model_utils.py
huzecong: This is the relative attention module as in TransformerXL, with adaptations for XLNet (segment embeddings and two-stream attention).

The difference between this and the vanilla multi-head attention is mainly two-fold:
1) it uses relative position embeddings (though in current implementation it's not computed in this module);
2) it attends not only to previous hidden states (`states_h`) but also to (detached) hidden states from previous batches (`memory`).

If necessary we could probably add a version without XLNet-specific adaptations to Texar modules.

texar/modules/pretrained/xlnet_base.py
gpengzhi: Because I here want to separate model downloading logic from the actual model building logic. What's more, there might be some modified version of the encoder in the future, and they can share the same base class. @AvinashBukkittu

texar/modules/regressors/xlnet_regressor.py
ZhitingHu: Using a module named `*Classifier` for regression is a bit weird. Can we do sth like:
```
class XLNetClassifier:
    def __init__(...):
        ... = XLNetEncoder(...)
        a_shared_function_building_output_layers(...)
        ... classifier_stuff ...

class XLNetRegressor:
    def __init__(...):
        ... = XLNetEncoder(...)
        a_shared_function_building_output_layers(...)
        ... regressor_stuff ...
```

Or, can we have sth like the previous `XLNetSummary` intermediate class, which is not exposed to users?

Can we allow users to config the output layers (e.g., number of layers, layer sizes, etc)? Is it necessary to allow it?

texar/modules/classifiers/xlnet_classifier.py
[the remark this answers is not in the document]
ZhitingHu: Then let's use the previous terminology and hparam structure

texar/modules/pretrained/xlnet_base.py
AvinashBukkittu: Hmm. What seems weird to me is that neither `XLNetBase` nor `XLNetEncoder` inherits from `EncoderBase` (although I know `EncoderBase` is an empty class essentially.) This is unlike other encoders we have in `texar`. At this point, I am thinking to have `XLNetEncoder` inherit from `XLNetBase` and `EncoderBase` (in this way we can still separate the logic of downloading the model and building the model.)

texar/modules/pretrained/xlnet_model_utils.py
ZhitingHu: `SinusoidsPositionEmbedder` is by @haoransh, can you confirm with him?

texar/modules/pretrained/xlnet_base.py
gpengzhi: What is the benefit to inherit from `EncoderBase`? @AvinashBukkittu

texar/modules/pretrained/xlnet_model_utils.py
huzecong: It seems to be because that was how `tensor2tensor` had implemented it. I would prefer changing our code to match the paper description. This shouldn't affect anything because the only other place we used `SinusoidsPositionalEmbedder` is in the seq2seq Transformer example, where we don't need to load existing models.

texar/modules/pretrained/xlnet_base.py
huzecong: I think the benefit would be a clear inheritance hierarchy, and we can check whether something is a decoder by `isinstance(x, EncoderBase)`.
This might require us to redesign constructor interfaces for `DecoderBase`, `EncoderBase`, and maybe `ClassifierBase`, so they allow cooperative inheritance.

AvinashBukkittu: If in the future, we want to add some additional properties to all the encoders, we can do it at one place i.e `EncoderBase`. Currently it is not possible for `XLNetEncoder` and `BertEncoder` as they don't inherit from the base class.

texar/modules/classifiers/xlnet_classifier.py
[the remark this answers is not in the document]
gpengzhi: Yes, I think we here should use `clas_strategy` instead of `summary_type`.

texar/modules/regressors/xlnet_regressor.py
gpengzhi: We currently only support one hidden-to-logits layer, and it is pretty straightforward to support multi-layer feature. But I am not sure if this is a helpful feature. Maybe we can make it a future enhancement?

huzecong: I don't really think it's necessary to add this feature. These classifiers are simply wrappers anyway, if the user needs this they can easily write their own classifiers.

texar/modules/pretrained/xlnet_base.py
gpengzhi: Yes, I think we should have a `PretrainedBase` instead of the different base classes for different pre-trained models.

texar/modules/pretrained/xlnet_model_utils.py
ZhitingHu: Would this affect the transformer performance by any chance? Do you think it's necessary to have an option in `hparams` so that users can choose `num_timescales - 1` if it's used for transformer?

Anyway, this embedder should be deleted and use the `SinusoidsPositionEmbedder` directly

texar/modules/pretrained/xlnet_base.py
ZhitingHu: I agree that all encoders should inherit `EncoderBase`

For XLNet, now `XLNetDecoder` inherits `XLNetEncoder`. Is it a good structure? Or perhaps a better way is to put shared things to `XLNetBase`, and both `XLNetDecoder` and `XLNetEncoder` inherit it?

texar/modules/pretrained/xlnet_model_utils.py
ZhitingHu: I think `Segment embedding` should not be included in this module. Like in BERT, all embeddings, including word embedding, segment embedding, and position embedding, are separated from `MultiheadAttentionEncoder`. They are aggregated externally as a single embedding before feeding to the encoder. Can we do the same factorization here?

For `two-stream attention`, we can probably have an option to toggle b/w single-stream or two-stream? Or at this point we can simply have a two-stream version for XLNet

Let's make the module a texar module

texar/modules/pretrained/xlnet_base.py
gpengzhi: Structure-wise, `XLNetDecoder` and `XLNetEncoder` are pretty similar. `XLNetDecoder` just has one more bias parameter. Here are my thoughts on the `pretrained` modules:

We have a general `PretrainedBase` and multiple utils files including `pretrained_utils.py`, `bert_utils.py`, `gpt2_utils.py`, and `xlnet_utils.py` (maybe also `xlnet_model_utils.py`). `pretrained_utils.py` contains some shared functions used in `xxx_utils.py`. `PretrainedBase` will call the corresponding `util` function according to the class name (Bertxxx, GPT2xxx, XLNetxxx), and every pretrained_encoder class will inherits `PretrainedBase`.

huzecong: I think calling based on the class name is not a good design pattern. We can add a class attribute for each of the pretrained models, e.g. `BertEncoder.model_name = "BERT`, and `PretrainedBase` can lookup required functions based on that.

BTW, I think `BertXXX` should be renamed to `BERTXXX`.

texar/modules/pretrained/xlnet_utils.py
gpengzhi: I prefer to keep these functions (`load_pretrained_bert`, `load_pretrained_gpt2`, and `load_pretrained_xlnet`) since there are some slight differences (mainly depending on how the model authors store the checkpoint files) among these functions.

texar/modules/encoders/gpt2_encoder.py
```diff
@@ -62,6 +64,8 @@ def __init__(self,
                  cache_dir: Optional[str] = None,
                  hparams=None):
 
+        self.model_name = "GPT2"
```
huzecong: Let's make it a class attribute so we can access `model_name` without having to instantiate the model:
```python
class GPT2Encoder(...):
	model_name = "GPT2"

	def __init__(...): ...
```

texar/modules/pretrained/xlnet_model_utils.py
huzecong: The difference here is that positional embeddings and segment embeddings are directly fed into each layer, while word embeddings are only used as the input to the first layer. So this module must take pos & seg embeddings as arguments.

Now two-stream attention is disabled when `states_g` is `None`.

I just think that this is a very model-specific implementation and might not be useful elsewhere. I'm more in favor of adding a general relative attention module (as in TransformerXL) to Texar modules, and keep this as is.

texar/modules/pretrained/xlnet_model_utils.py
huzecong: I don't really think so, as long as the model is trained from scratch.

I think the implementation here (which I adapted from the TransformerXL repo) is much simpler (and possibly faster) so maybe we can change `SinusoidsPositionEmbedder` to this implementation.

PR #105, opened 2019-07-16
Resolve #90 Integrate bleu_tool.py into texar.evals

gpengzhi:
#90

PR #109, opened 2019-07-20
Minor fix in BLEU

### LINKED ISSUE #90: Integrate bleu_tool.py into texar.evals

gpengzhi:
Resolve comments in https://github.com/asyml/texar-pytorch/pull/105

Resolve #90


examples/transformer/bleu_main.py
```diff
@@ -29,7 +29,7 @@
 
 from argparse import ArgumentParser
 
-from texar.evals.bleu_tool import bleu_wrapper
+from texar.evals.bleu_transformer import bleu_wrapper
```
ZhitingHu: Do we still have `bleu_wrapper`?

texar/evals/bleu_transformer.py
```diff
@@ -185,14 +194,18 @@ def bleu_tokenize(string: str) -> List[str]:
...
 
 
-def bleu_wrapper(ref_filename: str,
-                 hyp_filename: str,
-                 case_sensitive: bool = False) -> float:
+def file_bleu(ref_filename: str,
+              hyp_filename: str,
+              bleu_version: str = "corpus_bleu_transformer",
+              case_sensitive: bool = False) -> float:
     r"""Compute BLEU for two files (reference and hypothesis translation).
```
ZhitingHu: If `file_bleu` and `bleu_tokenize` are not specific for the bleu-transformer version, they can be moved to another file (either `bleu.py` or a new file)

gpengzhi: I prefer to keep these two functions in `bleu_transformer` for now. Currently `bleu_tokenize` is just a special case of the original implementation of the tokenization in `multi-bleu.perl` (This is also mentioned in `doc` of `bleu_tokenize`).  I rename `bleu_tokenize` to `bleu_tokenize_transformer`. After we implement the general `bleu_tokenize`, we could move /delete these two function to the correct folder.

ZhitingHu: Ok. I guess `bleu_transformer_tokenize` could prbably a better name?

PR #110, opened 2019-07-20
Add raw_output to TransformerDecoderOutput

wwt17:
Probably some users (like me) may want the output state before `output_layer` in `TransformerDecoderOutput`, like `cell_output` in `BasicRNNDecoderOutput`. I named it `raw_output`. What do you think of this name?
This PR is also applicable to the Tensorflow version.
Also, I corrected the doc of input of `GPT2Decoder`.


gpengzhi:
@wwt17 Thank you for correcting the `doc` issue in `GPT2Decoder`. What is your thought on this interface change? @ZhitingHu @huzecong

huzecong:
Sorry for accidentally closing it...
Thanks for the PR! While I understand the importance of such functionality, I would prefer not changing the interface because: 1) it breaks backwards compatibility; 2) it would require changing every other decoder classes for consistency.

There are ways to workaround this in the current framework. For instance, you can pass `tx.core.identity` as `output_layer` and construct your own output layer somewhere else. Would this solution satisfy your needs?

wwt17:
Thanks! I agree with @huzecong 's points. I do have implemented a workaround in my codebase. But I think it can make the interface more consistent with RNNs, and in most cases, it won't require too much change if accesses through attributes are used. Would it be possible to add this feature into the next big release?
Anyway, you can close this issue if you want. It is not a too significant change.

huzecong:
I think this is something we could have, but we might need some time to redesign the interfaces. For example, I think the type of `raw_outputs` (or maybe `state` is more appropriate?) should be `State`, which is `HiddenState` (i.e. `Tuple[Tensor, Tensor]`) for RNN decoders. However, for Transformer decoder `State` is `Cache`, which includes per-layer memories and an attentional bias, which creates inconsistencies. Nonetheless, this would be a chance to refactor our decoder implementations.

I am going to close this PR for now and open an enhancement issue instead. Thanks for your contributions! @gpengzhi Could you make a separate PR to fix GPT2 docstrings, while referencing this PR?

PR #116, opened 2019-07-23
Add vae_text example

TomNong:
* Port vae_text from texar-TF.

* Add external distribution
  MultivariateNormalDiag.

* Add preprocessing for data batch.

* Modify None checking condition for `initial_state` in 
   `RNNDecoderBase`.

* Modify `max_pos` for config_trans_yahoo.py.

* Modify connectors `mlp` function.

* Polish code.


examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,463 @@
...
+        # decoder
+        if config.decoder_type == "lstm":
+            # concat latent variable to input at every time step
+            latent_z = torch.unsqueeze(latent_z, 1)
+
+            latent_z = latent_z.repeat([1, output_embed.size(1), 1])
+            output_embed = torch.cat([output_embed, latent_z], 2)
+
+            train_helper = self.decoder.create_helper(
+                decoding_strategy='train_greedy')
```
huzecong: There's no need to create the train helper every time. It will be cached after it is created for the first time. Actually, since by default the `train_greedy` helper is used for training, there's no need to explicitly create it.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,463 @@
...
+        rc_loss = tx.losses.sequence_sparse_softmax_cross_entropy(
+            labels=text_ids[:, 1:],
+            logits=logits,
+            sequence_length=data_batch["length"] - 1)
+
+        nll = rc_loss + kl_weight * kl_loss
+
+        fetches = {
+            "nll": nll,
+            "kl_loss": kl_loss.detach(),
```
huzecong: Don't detach variables here unless it is necessary for the model. If variables are detached just because they needed to be summed up, call `.item()` in the training loop.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,463 @@
...
+    if not os.path.exists(save_dir):
+        os.makedirs(save_dir)
+
+    suffix = f"{config.dataset}_{config.decoder_type}Decoder.ckpt"
+
+    save_path = os.path.join(save_dir, suffix)
+
+    # KL term annealing rate
+    anneal_r = 1.0 / (config.kl_anneal_hparams["warm_up"] *
+        (train_data.__len__() / config.batch_size))
```
huzecong: Use `len(x)` instead of `x.__len__()`.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,463 @@
...
+    optimizer = tx.core.get_optimizer(params=model.parameters(),
+                                    hparams=config.opt_hparams)
+
+    def _run_epoch(epoch, mode, display=10):
+        if mode == 'train':
+            iterator.switch_to_train_data()
+        elif mode == 'valid':
+            iterator.switch_to_val_data()
+        elif mode == 'test':
+            iterator.switch_to_test_data()
```
huzecong: Let's just use `DataIterator` and call `iterator.switch_to_dataset(mode)` here.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,463 @@
...
+        kl_loss_ = rc_loss_ = 0.
+
+        for batch in iterator:
+            if mode == 'train':
+                opt_vars["kl_weight"] = min(
+                    1.0, opt_vars["kl_weight"] + anneal_r)
+
+                kl_weight_ = opt_vars["kl_weight"]
+            else:
+                kl_weight_ = 1.0
```
huzecong: Move this outside the loop.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,463 @@
...
+
+            mode = mode
```
huzecong: What's the point of this statement?

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,463 @@
...
+
+            if mode == "train":
+                adjust_learning_rate(optimizer, opt_vars["learning_rate"])
```
huzecong: Let's use an LR scheduler instead.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,463 @@
...
+            iterator.switch_to_test_data()
+        if mode == 'train':
+            model.train()
+        else:
+            model.eval()
+        step = 0
+        start_time = time.time()
+        num_words = num_sents = 0
+        nll_ = 0.
+        kl_loss_ = rc_loss_ = 0.
```
huzecong: Can we use `AverageRecorder` instead of manually keep track of these?

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,463 @@
...
+
+            batch_size_ = len(fetches_["lengths"])
+            num_sents += batch_size_
+            num_words += sum(fetches_["lengths"])
+            nll_ += fetches_["nll"].item() * batch_size_
+            kl_loss_ += fetches_["kl_loss"] * batch_size_
+            rc_loss_ += fetches_["rc_loss"] * batch_size_
+
+            if step % display == 0 and mode == 'train':
+                nll = nll_ / float(num_sents)
```
huzecong: Just as a side note: no need to call `float` here. Division by a single slash (`/`) always returns floating point numbers. Double slash (`//`) is integer division.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,463 @@
...
+        KL = kl_loss_ / float(num_sents)
+        rc = rc_loss_ / float(num_sents)
+        log_ppl = nll_ / float(num_words)
+        ppl = np.exp(log_ppl)
+        print(f"\n{mode}: epoch {epoch}, nll {nll:.4f}, KL {KL:.4f}, "
+              f"rc {rc:.4f}, log_ppl {log_ppl:.4f}, ppl {ppl:.4f}")
+        return nll, ppl
+
+    @torch.no_grad()
+    def _generate(fname=None):
```
huzecong: Call this `filename` or `path` instead of ambiguous `fname`.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,463 @@
...
+        ckpt = torch.load(args.model)
+        model.load_state_dict(ckpt['model'])
+        model.eval()
+
+        batch_size = train_data.batch_size
+
+        dst = MultivariateNormalDiag(
```
huzecong: Isn't this what `model.forward` does? Why copy it here?

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,463 @@
...
+                memory_sequence_length=torch.ones(dcdr_states.size(0)),
+                max_decoding_length=100,
+                decoding_strategy='infer_sample',
+                embedding=_embedding_fn,
+                start_tokens=start_tokens,
+                end_token=end_token)
+
+        sample_tokens = vocab.map_ids_to_tokens_py(outputs.sample_id.cpu())
+
+        sample_tokens_ = sample_tokens
```
huzecong: This statement is pointless.

texar/modules/decoders/rnn_decoder_base.py
```diff
@@ -201,7 +201,10 @@ def initialize(self, helper: Helper, inputs: Optional[torch.Tensor],
             -> Tuple[torch.ByteTensor, torch.Tensor, Optional[State]]:
         initial_finished, initial_inputs = helper.initialize(
             inputs, sequence_length)
-        state = initial_state or self._cell.init_batch()
+        if initial_state is None:
+            state = self._cell.init_batch()
+        else:
+            state = initial_state
```
huzecong: Why is this change required?

texar/custom/__init__.py
```diff
@@ -4,3 +4,4 @@
 
 from texar.custom.activation import *
 from texar.custom.initializers import *
+from texar.custom.distributions import *
```
huzecong: I believe you forgot to commit `distributions.py`.

texar/modules/connectors/connectors.py
```diff
@@ -37,6 +37,8 @@
     "MLPTransformConnector",
     "ReparameterizedStochasticConnector",
     "StochasticConnector",
+    "sum_output_size",
+    "mlp_transform"
```
huzecong: I think these two methods are intended for internal usage only. Are they used somewhere else?

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,463 @@
...
+To run:
+
+$ python vae_train.py
+
+Hyperparameters and data path may be specified in config_trans.py
+
+"""
+
+# pylint: disable=invalid-name, no-member, too-many-locals
+# pylint: disable=too-many-branches, too-many-statements
```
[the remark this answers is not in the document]
TomNong: Got it.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,463 @@
...
+
+import os
+import sys
+import time
+import argparse
+import importlib
+from io import open
```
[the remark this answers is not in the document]
TomNong: Fixed.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,463 @@
...
+    return torch.sum(kl_cost)
+
+
+def adjust_learning_rate(optimizer, lr):
+    """Sets the learning rate to the initial LR decayed by 10 every 30 epochs"""
+    for param_group in optimizer.param_groups:
+        param_group['lr'] = lr
+
+
+class Vae(nn.Module):
```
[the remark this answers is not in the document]
TomNong: Fixed.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,463 @@
...
+    def __init__(self, train_data):
```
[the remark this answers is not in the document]
TomNong: Got it

examples/vae_text/vae_train.py
TomNong: Yes, it is similar to what `model.forward` does, do we want to implement L344-L392 into `model.forward`?

examples/vae_text/vae_train.py
TomNong: This one is a typo.

texar/modules/decoders/rnn_decoder_base.py
TomNong: It should fix the `None` confusion.

texar/custom/__init__.py
TomNong: Indeed! Thanks

texar/modules/connectors/connectors.py
TomNong: I see, I think we can use simpler methods instead, so to keep them internal.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,463 @@
...
+
+        if config.decoder_type == "lstm":
+            output_embed = output_w_embed
+
+        elif config.decoder_type == 'transformer':
+
+            batch_size = text_ids.size(0)
+            max_seq_len = text_ids.size(1) - 1
+            batch_max_seq_len = torch.ones(
+                batch_size, dtype=torch.long, device=device) * max_seq_len
```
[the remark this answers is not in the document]
TomNong: I see, should we use `torch.full` for all `torch.ones`?

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,463 @@
...
+        self.mlp_linear_layer = nn.Linear(
+            config.batch_size,
+            tx.modules.connectors.sum_output_size(decoder_initial_state_size))
+
+    def forward(self, data_batch, kl_weight):
+        # encoder -> connector -> decoder
+
+        tmp_batch = []
+        for i, data in enumerate(data_batch["text_ids"]):
+            filter_data = data[data != 3]
```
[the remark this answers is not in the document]
TomNong: I see, I should create a `VAEData` class

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,463 @@
...
+        self.encoder = tx.modules.UnidirectionalRNNEncoder(
+            input_size=self.encoder_w_embedder.dim,
+            hparams={
+                "rnn_cell": config.enc_cell_hparams,
+            })
+
+        self.decoder_w_embedder = tx.modules.WordEmbedder(
+            vocab_size=train_data.vocab.size, hparams=config.dec_emb_hparams)
+
+        if config.decoder_type == "lstm":
```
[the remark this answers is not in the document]
TomNong: Got it

examples/vae_text/vae_train.py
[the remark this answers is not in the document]
huzecong: No. `torch.full` should replace current usages of `torch.ones(...) * some_number`.

examples/vae_text/vae_train.py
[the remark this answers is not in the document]
huzecong: That would be preferable. `VAEData` can subclass `MonoTextData` so it suffices to just override `collate`.

examples/vae_text/vae_train.py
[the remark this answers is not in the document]
TomNong: Now config is passed as argument.

examples/vae_text/vae_train.py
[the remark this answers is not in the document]
TomNong: VAEData is added.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,463 @@
...
+                    help="Generation output path.")
+
+args = parser.parse_args()
+
+config = importlib.import_module(args.config)
+
+device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+
+
+def kl_dvg(means, logvars):
```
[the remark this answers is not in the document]
TomNong: Checked.

examples/vae_text/vae_train.py
TomNong: Corrected.

examples/vae_text/vae_train.py
TomNong: `DataIterator` is implemented.

examples/vae_text/vae_train.py
TomNong: We may need to keep the training case.

examples/vae_text/vae_train.py
TomNong: Removed.

examples/vae_text/vae_train.py
TomNong: Corrected.

examples/vae_text/vae_train.py
TomNong: Corrected, thanks for pointing out!

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,518 @@
...
+                rc = avg_rec.avg(2)
+                log_ppl = nll_ / num_words
+                ppl = np.exp(log_ppl)
+                time_cost = time.time() - start_time
+
+                print(f"{mode}: epoch {epoch}, step {step}, nll {nll:.4f}, "
+                      f"klw {klw:.4f}, KL {KL:.4f}, rc {rc:.4f}, "
+                      f"log_ppl {log_ppl:.4f}, ppl {ppl:.4f}, "
+                      f"time_cost {time_cost:.1f}")
+                sys.stdout.flush()
```
huzecong: Use `print(..., flush=True)`.

examples/vae_text/vae_train.py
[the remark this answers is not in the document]
TomNong: Replaced.

examples/vae_text/vae_train.py
TomNong: `AverageRecorder` added for some variables.

examples/vae_text/vae_train.py
TomNong: Corrected.

examples/vae_text/vae_train.py
TomNong: Corrected.

texar/modules/connectors/connectors.py
```diff
@@ -96,8 +96,8 @@ def _sum_output_size(output_size: OutputSize) -> int:
             size_list[i] = np.prod([dim for dim in shape])
     else:
         size_list = flat_output_size
-    sum_output_size = sum(size_list)
-    return sum_output_size
+    _sum_output_size_ = sum(size_list)
+    return _sum_output_size_
```
huzecong: Why such naming?

examples/vae_text/vae_train.py
TomNong: New method `decode` is implemented for the decoder parts.

texar/custom/__init__.py
TomNong: Corrected.

texar/modules/connectors/connectors.py
TomNong: These two functions are restored.

texar/modules/connectors/connectors.py
TomNong: Oh, just notice it.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,518 @@
...
+            labels=text_ids[:, 1:],
+            logits=logits,
+            sequence_length=seq_lengths)
+
+        nll = rc_loss + kl_weight * kl_loss
+
+        fetches = {
+            "nll": nll,
+            "kl_loss": kl_loss.item(),
+            "rc_loss": rc_loss.item(),
```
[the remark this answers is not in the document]
TomNong: Got it.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,518 @@
...
+        }
+
+        return fetches
+
+    def decode(self,
+        dcdr_states: Tensor,
+        latent_z: Tensor,
+        output_embed: Optional[Tensor] = None,
+        seq_lengths: Optional[Tensor] = None,
+        is_generate: bool = False) -> \
```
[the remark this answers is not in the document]
TomNong: Sounds good.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,518 @@
...
+
+    # KL term annealing rate
+    anneal_r = 1.0 / (config.kl_anneal_hparams["warm_up"] *
+        (len(train_data) / config.batch_size))
+
+    vocab = train_data.vocab
+    model = VAE(
+        train_data.vocab.size,
+        vocab.bos_token_id.item(),
+        vocab.eos_token_id.item(), config)
```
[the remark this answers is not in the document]
TomNong: Got it, thanks!

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,518 @@
...
+    scheduler = ExponentialLR(optimizer, decay_factor)
+
+    def _run_epoch(epoch: int, mode: str, display: int = 10) -> \
+        Tuple[Tensor, float]:
+        if mode == 'train':
+            iterator.switch_to_dataset("train")
+        elif mode == 'valid':
+            iterator.switch_to_dataset("valid")
+        elif mode == 'test':
+            iterator.switch_to_dataset("test")
```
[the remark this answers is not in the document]
TomNong: Ah, forgive me.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,518 @@
...
+            kl_weight_ = 1.0
+        step = 0
+        start_time = time.time()
+        num_words = 0
+        nll_ = 0.
+
+        avg_rec = tx.utils.AverageRecorder()
+        for batch in iterator:
+
+            fetches_ = model(batch, kl_weight_)
```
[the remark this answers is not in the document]
TomNong: Got that.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,518 @@
...
+    kl_cost = -0.5 * (logvars - means**2 -
+                      torch.exp(logvars) + 1.0)
+    kl_cost = torch.mean(kl_cost, 0)
+
+    return torch.sum(kl_cost)
+
+
+class VAE(nn.Module):
+    def __init__(self,
+        vocab_size: int, bos_token_id: int,
```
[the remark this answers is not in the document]
TomNong: Got it.

examples/vae_text/vae_train.py
TomNong: Let me know if `decode` need any improvements, thanks

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,518 @@
...
+                    config_model.batch_size),
+                hparams={"rnn_cell": config_model.dec_cell_hparams})
+            decoder_initial_state_size = (self.decoder.cell.hidden_size,
+                                          self.decoder.cell.hidden_size)
+            sum_state_size = sum(decoder_initial_state_size)
+
+        elif config_model.decoder_type == 'transformer':
+            decoder_initial_state_size = torch.Size(
+                [1, config_model.dec_emb_hparams["dim"]])
+            sum_state_size = np.prod(decoder_initial_state_size)
```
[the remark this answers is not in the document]
TomNong: Fixed.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,518 @@
...
+
+        if config_model.decoder_type == "lstm":
+            self.decoder = tx.modules.BasicRNNDecoder(
+                vocab_size=vocab_size,
+                input_size=(self.decoder_w_embedder.dim +
+                    config_model.batch_size),
+                hparams={"rnn_cell": config_model.dec_cell_hparams})
+            decoder_initial_state_size = (self.decoder.cell.hidden_size,
+                                          self.decoder.cell.hidden_size)
+            sum_state_size = sum(decoder_initial_state_size)
```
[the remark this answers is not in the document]
TomNong: That is neat.

examples/vae_text/vae_train.py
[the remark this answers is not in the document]
TomNong: It is fixed now.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,518 @@
...
+            if mode == "train":
+                opt_vars["kl_weight"] = min(
+                    1.0, opt_vars["kl_weight"] + anneal_r)
+                kl_weight_ = opt_vars["kl_weight"]
+                fetches_["nll"].backward()
+                optimizer.step()
+                optimizer.zero_grad()
+
+            batch_size_ = len(fetches_["lengths"])
+            num_words += sum(fetches_["lengths"]).item()
```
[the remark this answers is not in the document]
TomNong: Fixed.

examples/vae_text/vae_train.py
TomNong: Fixed.

texar/modules/connectors/connectors.py
TomNong: Assigned with the new names.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,518 @@
...
+parser.add_argument('--out',
+                    type=str,
+                    default=None,
+                    help="Generation output path.")
+
+args = parser.parse_args()
+
+config: Any = importlib.import_module(args.config)
+
+device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
```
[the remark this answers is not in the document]
TomNong: Fixed.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,518 @@
...
+
+        self.connector_mlp = tx.modules.MLPTransformConnector(
+            config_model.latent_dims * 2,
+            linear_layer_dim=self.encoder.cell.hidden_size * 2)
+
+        self.mlp_linear_layer = nn.Linear(
+            config_model.latent_dims,
+            sum_state_size)
+
+    def forward(self, data_batch: Dict, kl_weight: float) -> Dict:
```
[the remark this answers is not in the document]
TomNong: Fixed.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,518 @@
...
+
+        dst = MultivariateNormalDiag(
+            loc=mean,
+            scale_diag=torch.exp(0.5 * logvar))
+
+        latent_z = dst.rsample()
+
+        fc_output = self.mlp_linear_layer(latent_z)
+        if self._config.decoder_type == "lstm":
+            dcdr_states = torch.split(
```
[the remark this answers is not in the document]
TomNong: Corrected.

examples/vae_text/vae_train.py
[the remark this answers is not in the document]
TomNong: Here is fixed, should we correct this format for other part in texar?

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,518 @@
...
+        ]
+
+        text_ids = torch.from_numpy(text_ids).to(device=self.device)
+        lengths = torch.tensor(lengths, dtype=torch.long, device=self.device)
+        tmp_batch = []
+        for i, data in enumerate(text_ids):
+            filter_data = data[data != BLANK_ID]
+            delta = data.size(0) - filter_data.size(0)
+            tmp_batch.append(filter_data)
+            lengths[i] -= delta
```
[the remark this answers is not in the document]
TomNong: `process` is added, let me know if there can be improvements.

examples/vae_text/vae_train.py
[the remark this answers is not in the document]
TomNong: Fixed.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,518 @@
...
+        self.mlp_linear_layer = nn.Linear(
+            config_model.latent_dims,
+            sum_state_size)
+
+    def forward(self, data_batch: Dict, kl_weight: float) -> Dict:
+        # encoder -> connector -> decoder
+        text_ids = data_batch["text_ids"]
+        input_embed = self.encoder_w_embedder(text_ids)
+        output_w_embed = self.decoder_w_embedder(text_ids[:, :-1])
+        _, ecdr_states = self.encoder(
```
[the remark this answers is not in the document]
TomNong: It was TF legacy, Fixed.

examples/vae_text/vae_train.py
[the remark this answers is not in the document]
TomNong: Fixed.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,463 @@
...
+            self.decoder = tx.modules.TransformerDecoder(
+                # tie word embedding with output layer
+                output_layer=self.decoder_w_embedder.embedding,
+                hparams=config.trans_hparams)
+        else:
+            raise NotImplementedError
+
+        self.decoder_initial_state_size = decoder_initial_state_size
+
+        self.connector_mlp = tx.modules.MLPTransformConnector(
```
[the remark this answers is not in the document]
TomNong: This one can follow the discussion.

examples/vae_text/vae_train.py
TomNong: This one is fixed in the latest commit.

examples/vae_text/vae_train.py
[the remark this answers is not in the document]
TomNong: This is fixed in the later commits.

examples/vae_text/vae_train.py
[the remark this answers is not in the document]
huzecong: I don't think you got what I mean... I meant `sum_state_size = self.decoder.cell.hidden_size * 2`.

examples/vae_text/vae_train.py
```diff
@@ -107,14 +103,13 @@ def __init__(self,
                 input_size=(self.decoder_w_embedder.dim +
```
huzecong: I don't think this is right. Now decoders take either `token_embedder` or `token_pos_embedder` as argument (which can be `nn.Module` or a function) in their constructors. Since you also need to concatenate `latent_z`, I would suggest storing `latent_z` as an attribute and writing a method as `token_embedder`.

examples/vae_text/vae_train.py
```diff
@@ -107,14 +103,13 @@ def __init__(self,
...
             sum_state_size = sum(decoder_initial_state_size)
 
         elif config_model.decoder_type == 'transformer':
             decoder_initial_state_size = torch.Size(
                 [1, config_model.dec_emb_hparams["dim"]])
-            sum_state_size = np.prod(decoder_initial_state_size)
+            sum_state_size = config_model.dec_emb_hparams["dim"]
             # position embedding
             self.decoder_p_embedder = tx.modules.SinusoidsPositionEmbedder(
                 position_size=config_model.max_pos,
```
huzecong: And also for line 118. You should create a function that generates word embeddings combined with positional embeddings, and pass that function as `token_pos_embedder`. I would suggest making that function a method of this `VAE` class.

huzecong: And in line 123, instead of raising `NotImplementedError`, let's use `ValueError` with an error message. `NotImplementedError` has a very specific use case: for abstract methods that ought to be overridden by subclasses.

examples/vae_text/vae_train.py
```diff
@@ -196,111 +193,82 @@ def forward(self, data_batch: Dict, kl_weight: float) -> Dict:
...
-                    memory=dcdr_states,
-                    memory_sequence_length=torch.ones(dcdr_states.size(0)),
-                    max_decoding_length=100,
-                    decoding_strategy='infer_sample',
-                    embedding=_embedding_fn,
-                    start_tokens=start_tokens,
-                    end_token=end_token)
+                output_embed = None
+            outputs = self.decoder(
+                inputs=output_embed,
```
huzecong: And here you would pass `text_ids` to `inputs`. No need to compute `output_embed` here.

examples/vae_text/vae_train.py
```diff
@@ -196,111 +193,82 @@ def forward(self, data_batch: Dict, kl_weight: float) -> Dict:
...
-            filter_data = data[data != BLANK_ID]
-            delta = data.size(0) - filter_data.size(0)
-            tmp_batch.append(filter_data)
-            lengths[i] -= delta
-        text_ids = torch.stack(tmp_batch).to(device)
-        batch = {self.text_id_name: text_ids,
-                 self.length_name: lengths}
-        return Batch(len(examples), batch=batch)
+    def process(self, raw_example: str) -> List[str]:
+        # Truncates sentences and appends BOS/EOS tokens.
```
huzecong: I'm confused. Isn't this method exactly the same as `MonoTextData.process`? What is this specific preprocessing step that you require for VAE?
And also, if you still need the preprocessing steps of `MonoTextData`, you can always call `super().process(raw_example)` in this method.

And by the way, after the latest PR is merged, the `raw_example` argument in `MonoTextData.process` will become the tokenized input, i.e. a list of `str`s. It shouldn't require much change on your side though.

examples/vae_text/vae_train.py
```diff
@@ -433,26 +401,52 @@ def _generate(filename: Optional[str] = None):
...
-        outputs = model.decode(
-            dcdr_states, latent_z, output_embed=None,
-            seq_lengths=None, is_generate=True)
+        if config.decoder_type == "lstm":
+            def _cat_embedder(ids):
+                """Concatenates latent variable to input word embeddings
+                """
+                embedding = model.decoder_w_embedder(
+                    ids.to(device))
+                return torch.cat([embedding, latent_z], 1)
```
huzecong: Based on the current implementation, can I assume you only concatenated the `latent_z`'s during generation but not training? This would be incorrect and result in poor performance.

examples/vae_text/vae_train.py
TomNong: Got it, after implementing `token_pos_embedder`.

examples/vae_text/vae_train.py
TomNong: Got it.

examples/vae_text/vae_train.py
TomNong: Got it, implemented in the `token_embedder`.

examples/vae_text/vae_train.py
TomNong: Got it.

examples/vae_text/vae_train.py
```diff
@@ -154,40 +154,38 @@ def forward(self, data_batch: tx.data.Batch,
...
-                embedding=_embedding_fn,
                 start_tokens=start_tokens,
                 end_token=end_token,)
+            latent_z = torch.unsqueeze(latent_z, 0)
+            latent_z = latent_z.repeat([text_ids.size(1), 1, 1])
         else:
             decoder_states = torch.reshape(
                 fc_output, (-1, ) + self.decoder_initial_state_size)
             helper = None
+            text_ids = text_ids[:, :-1]
```
huzecong: Have you tested both RNN & Transformer models? Why is this in the `else` branch?

examples/vae_text/vae_train.py
```diff
@@ -202,14 +200,36 @@ def _embedding_fn(token):
...
+        output_embed = output_w_embed + output_p_embed
+        return output_embed
+
     def decode(self,
                decoder_states: Union[Tensor, Tuple],
                helper: Optional[Helper],
+               latent_z: Optional[Tensor] = None,
                seq_lengths: Optional[Tensor] = None,
                max_decoding_length: int = None,
                text_ids: torch.LongTensor = None) \
```
huzecong: These type annotations will not pass type-check.

examples/vae_text/vae_train.py
```diff
@@ -243,23 +250,11 @@ def decode(self,
...
-        if self._hparams.dataset["eos_token"] != '':
-            words.append(self._hparams.dataset["eos_token"])
-
-        # Apply the "other transformations".
-        for transform in self._other_transforms:
-            words = transform(words)
-
+        raw_example = super().process(raw_example)
+        words = " ".join(raw_example[1:-1]).strip().split(self._delimiter)
+        words = [raw_example[0]] + words + [raw_example[-1]]
```
huzecong: I don't understand what you're doing here. How will `words` differ from `raw_example`?

huzecong: Also, #118 is already fixed.

examples/vae_text/vae_train.py
```diff
@@ -408,34 +401,25 @@ def _generate(start_tokens: torch.LongTensor,
                 fc_output, (-1, ) + model.decoder_initial_state_size)
```
huzecong: Have you realized that the logic for computing `decoder_states` is exactly the same as in `forward`? Can we move it to `decode`?

examples/vae_text/vae_train.py
```diff
@@ -99,12 +97,13 @@ def __init__(self,
 
         if config_model.decoder_type == "lstm":
             self.decoder = tx.modules.BasicRNNDecoder(
-                vocab_size=vocab_size,
                 input_size=(self.decoder_w_embedder.dim +
                     config_model.batch_size),
+                vocab_size=vocab_size,
+                token_embedder=self._embed_fn_rnn,
                 hparams={"rnn_cell": config_model.dec_cell_hparams})
             decoder_initial_state_size = (self.decoder.cell.hidden_size,) * 2
```
huzecong: I don't think you need this `decoder_initial_state_size` here. In the code below, whenever you used this variable, you still check for `decoder_type`. Why not just directly use what was assigned to this variable in each case?

TomNong:
@huzecong The training & generation results are as good as past version.

examples/vae_text/vae_train.py
TomNong: You are right, that can work. Just fix it.

examples/vae_text/vae_train.py
TomNong: Yes, I'll put it into `decode`.

examples/vae_text/vae_train.py
TomNong: Yes, got it.

examples/vae_text/vae_train.py
TomNong: I see, will correct `max_decoding_length ` and `text_ids `.

TomNong:
@huzecong Got it, I will do it.

PR #124, opened 2019-07-26
nn.Identity bugfix

### LINKED ISSUE #123: AttributeError: module 'torch.nn' has no attribute 'Identity'
The error occurs when I am running the example of gpt-2.

python gpt2_generate_main.py --is_interactive \
--max_decoding_length=100 \
--temperature=0.7 \
--top_k=40

image

torch version: 1.0.0

gpengzhi:
Resolve #123


huzecong:
LGTM, but when did we introduce the `Identity` layer? I thought we had a functional `identity` also under `texar.core`, and in decoder we check whether `output_layer` is the functional `identity`?

gpengzhi:
I didn't realize that. But I think the logic is exactly the same here. Maybe both of them are OK for this PR?

texar/modules/decoders/decoder_base.py
```diff
@@ -65,6 +70,7 @@ def _make_output_layer(layer: Optional[Union[nn.Module, torch.Tensor]],
                 "wanted.")
         output_layer = nn.Linear(output_size, vocab_size, bias)
     elif torch.is_tensor(layer):
+        assert isinstance(layer, torch.Tensor)
```
huzecong: `torch.is_tensor` is actually just `isinstance(_, torch.Tensor)`. We can just replace this function call with `isinstance`.

PR #125, opened 2019-07-26
Polish XLNet example

gpengzhi:
Refactor xlnet example using the lastest xlnet module.


examples/xlnet/xlnet_classification_main.py
```diff
@@ -11,22 +11,32 @@
...
 import tqdm
 import sentencepiece as spm
+
 import texar as tx
 
-import xlnet
+# pylint: disable=wildcard-import
+
+from utils.classification import *
+from utils.glue import *
```
huzecong: Can we put data loaders and stuff under a separate package? It feels weird to import data loaders from `utils`.

examples/xlnet/xlnet_classification_main.py
```diff
@@ -55,9 +65,10 @@ def parse_args():
         help="Directory to save model checkpoints")
 
     parser.add_argument(
-        "--pretrained", type=str,
-        default="pretrained/xlnet_cased_L-24_H-1024_A-16/xlnet_model.ckpt",
-        help="Path to the pretrained XLNet model checkpoint")
+        "--pretrained_model_name", type=str,
```
huzecong: Let's use consistent naming styles. Now we're using `argument-names-like-this`, but not `like_this`.

examples/xlnet/xlnet_classification_main.py
```diff
@@ -110,22 +123,34 @@ def evaluate(model, iterator, is_regression: bool = False, print_fn=None,
...
-        args.task, cache_dir, args.max_seq_len, args.batch_size,
-        file_prefix=cache_prefix, eval_batch_size=args.eval_batch_size,
-        shuffle_buffer=None, device=device)
+    dataset.construct_dataset(task_processor,
+                              cache_dir,
+                              args.max_seq_len,
+                              tokenize_fn,
+                              file_prefix=cache_prefix)
+
+    datasets = dataset.load_datasets(args.task,
```
huzecong: Can we reformat this part of code to the previous style, i.e. wrap before the first argument? This would result in way fewer lines and make the code look less "sparse".

examples/xlnet/xlnet_generation_main.py
```diff
@@ -14,28 +14,22 @@
...
-                    default="pretrained/xlnet_cased_L-24_H-1024_A-16/"
-                            "xlnet_model.ckpt",
-                    help="XLNet pretrained model checkpoint. Ignored if "
-                         "'--checkpoint' is specified.")
-parser.add_argument('--pretrain_model_dir', type=str,
-                    default="pretrained/xlnet_cased_L-24_H-1024_A-16",
-                    help="The directory of pretrained model, for loading "
-                         "vocabulary, etc.")
+                    help="Checkpoint to load model weights from.")
+parser.add_argument("--pretrained_model_name", type=str,
```
huzecong: Same here. Avinash wrote this file so it didn't follow the same naming convention. Let's unify them.

examples/xlnet/README.md
```diff
@@ -301,11 +272,11 @@ module in other tasks.
 #### Use your own dataset by writing a custom data processor
 
 It is easy to adapt the code to fine-tune XLNet on your custom dataset. To do this, you will need to write a custom
-data processor inheriting `xlnet.data.DataProcessor`. For concrete examples, please refer to the built-in processors
-under `xlnet/data/classification.py` and `xlnet/data/glue.py`.
+data processor inheriting `utils.processor.DataProcessor`. For concrete examples, please refer to the built-in processors
+under `utils/classification.py` and `utils/glue.py`.
```
ZhitingHu: There are lots of files under `utils/`. Can we merge them? E.g., all data relevant stuff (data utils, processors, etc) can go into a single file. Make the file structure look as simple as possible

ZhitingHu: BTW, what're the Processors for? What's their relation with Texar data modules?

gpengzhi: Maybe we can have a different folder for data-related files? Processors here are just data loader for different datasets in GLUE. After we use the processor loads the dataset (.tsv file). We will use texar data module to convert those files into pickled files.

ZhitingHu: I see. At least all processors should be in a single file (now they are scattered in `processor`, `classification`, and `glue`, which looks complicated.).

huzecong: I'm actually in favor of separating data loaders by their tasks. We can group them together as a package.

ZhitingHu: It's also good to separate by tasks, instead of `classification.py` which is confusing. The file name should be the task name.

PR #133, opened 2019-07-30
Add vae_text example

TomNong:
* Port vae_text from texar-TF.

* Add external distribution
MultivariateNormalDiag.

* Add preprocessing for data batch.

* Modify None checking condition for initial_state in
RNNDecoderBase.

* Modify max_pos for config_trans_yahoo.py.

* Modify connectors mlp function.

* Refactor vae_text training & generation decoder.

* Refactor vae_text decoder embeddings.

* Refactor to import `texar.torch`.

* Polish code.

### DIFF
--- a/examples/vae_text/README.md
+++ b/examples/vae_text/README.md
@@ -1,0 +1,57 @@
+# Variational Autoencoder (VAE) for Text Generation
+
+This example builds a VAE for text generation, with an LSTM as encoder and an LSTM or [Transformer](https://arxiv.org/pdf/1706.03762.pdf) as decoder. Training is performed on the official PTB data and Yahoo data, respectively. 
+
+The VAE with LSTM decoder is first decribed in [(Bowman et al., 2015) Generating Sentences from a Continuous Space](https://arxiv.org/pdf/1511.06349.pdf)
+
+The Yahoo dataset is from [(Yang et al., 2017) Improved Variational Autoencoders for Text Modeling using Dilated Convolutions](https://arxiv.org/pdf/1702.08139.pdf), which is created by sampling 100k documents from the original Yahoo Answer data. The average document length is 78 and the vocab size is 200k. 
+
+## Data
+The datasets can be downloaded by running:
+```shell
+python prepare_data.py --data ptb
+python prepare_data.py --data yahoo
+```
+
+## Training
+Train with the following command:
+
+```shell
+python vae_train.py --config config_trans_ptb
+```
+
+Here:
+
+* `--config` specifies the config file to use, including model hyperparameters and data paths. We provide 4 config files:
+  - [config_lstm_ptb.py](./config_lstm_ptb.py): LSTM decoder, on the PTB data
+  - [config_lstm_yahoo.py](./config_lstm_yahoo.py): LSTM decoder, on the Yahoo data
+  - [config_trans_ptb.py](./config_trans_ptb.py): Transformer decoder, on the PTB data
+  - [config_trans_yahoo.py](./config_trans_yahoo.py): Transformer decoder, on the Yahoo data
+
+## Generation
+Generating sentences with pre-trained model can be performed with the following command:
+```shell
+python vae_train.py --config config_file --mode predict --model /path/to/model.ckpt --out /path/to/output
+```
+
+Here `--model` specifies the saved model checkpoint, which is saved in `./models/dataset_name/` at training time. For example, the model path is `./models/ptb/ptb_lstmDecoder.ckpt` when generating with a LSTM decoder trained on PTB dataset. Generated sentences will be written to standard output if `--out` is not specifcied.
+
+## Results
+
+### Language Modeling
+
+|Dataset    |Metrics   | VAE-LSTM |VAE-Transformer |
+|---------------|-------------|----------------|------------------------|
+|Yahoo | Test PPL<br>Test NLL | 68.11<br>337.13 |59.95<br>326.93|
+|PTB | Test PPL<br>Test NLL | 104.61<br>101.92 | 103.68<br>101.72 |
+
+### Generated Examples
+We show the generated examples with transformer as decoder trained  on PTB training data.
+
+|Examples|
+|:---------|
+|i 'm always looking at a level of \$ N to \$ N billion \<EOS\> |
+|after four years ago president bush has federal regulators decided to file financing for the waiver\<EOS\> |
+|the savings & loan association said total asset revenue was about \$ N billion compared with \$ N billion \<EOS\> |
+|the trend would seem to be effective \<EOS\> |
+|chicago city 's <unk> computer bank of britain posted a N N jump in third-quarter net income \<EOS\>|
--- a/examples/vae_text/README.md
+++ b/examples/vae_text/README.md
@@ -43,15 +43,15 @@ Here `--model` specifies the saved model checkpoint, which is saved in `./models
 |Dataset    |Metrics   | VAE-LSTM |VAE-Transformer |
 |---------------|-------------|----------------|------------------------|
 |Yahoo | Test PPL<br>Test NLL | 68.11<br>337.13 |59.95<br>326.93|
-|PTB | Test PPL<br>Test NLL | 104.61<br>101.92 | 103.68<br>101.72 |
+|PTB | Test PPL<br>Test NLL | 107.87<br>102.59 | 102.28<br>101.43 |
 
 ### Generated Examples
 We show the generated examples with transformer as decoder trained  on PTB training data.
 
 |Examples|
 |:---------|
-|i 'm always looking at a level of \$ N to \$ N billion \<EOS\> |
-|after four years ago president bush has federal regulators decided to file financing for the waiver\<EOS\> |
-|the savings & loan association said total asset revenue was about \$ N billion compared with \$ N billion \<EOS\> |
-|the trend would seem to be effective \<EOS\> |
-|chicago city 's <unk> computer bank of britain posted a N N jump in third-quarter net income \<EOS\>|
+|we 've been more quickly accomplished in our old and more than in the past is in these industries said steven <unk> publisher of <unk> <unk> inc. of the workstation market set crazy \<EOS\>|
+|chung <unk> 's proposal that saved the page in september and the ad pages \$ N from yale city with more than a small fee that tenants had been made by support in the past N years ago \<EOS\>|
+|the university of state commerce was <unk> from cooperation last year when it controls for cleaning up a bipartisan aid to moscow new arms-control <unk> <unk> \<EOS\>|
+|the <unk> maker general electric co. 's electronics and the <unk> for <unk> radar tools that included for space at krasnoyarsk \<EOS\>|
+|the company said its travel subsidiary has been operating for \$ N million \<EOS\>|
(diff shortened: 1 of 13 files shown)


examples/vae_text/README.md
```diff
@@ -0,0 +1,57 @@
...
+Here `--model` specifies the saved model checkpoint, which is saved in `./models/dataset_name/` at training time. For example, the model path is `./models/ptb/ptb_lstmDecoder.ckpt` when generating with a LSTM decoder trained on PTB dataset. Generated sentences will be written to standard output if `--out` is not specifcied.
+
+## Results
+
+### Language Modeling
+
+|Dataset    |Metrics   | VAE-LSTM |VAE-Transformer |
+|---------------|-------------|----------------|------------------------|
+|Yahoo | Test PPL<br>Test NLL | 68.11<br>337.13 |59.95<br>326.93|
+|PTB | Test PPL<br>Test NLL | 104.61<br>101.92 | 103.68<br>101.72 |
```
huzecong: Are these results from the TF version? Please update this with your results.

examples/vae_text/README.md
```diff
@@ -0,0 +1,57 @@
...
+### Generated Examples
+We show the generated examples with transformer as decoder trained  on PTB training data.
+
+|Examples|
+|:---------|
+|i 'm always looking at a level of \$ N to \$ N billion \<EOS\> |
+|after four years ago president bush has federal regulators decided to file financing for the waiver\<EOS\> |
+|the savings & loan association said total asset revenue was about \$ N billion compared with \$ N billion \<EOS\> |
+|the trend would seem to be effective \<EOS\> |
+|chicago city 's <unk> computer bank of britain posted a N N jump in third-quarter net income \<EOS\>|
```
huzecong: And your samples as well.

examples/vae_text/config_lstm_ptb.py
```diff
@@ -0,0 +1,138 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""VAE config.
+"""
+
+# pylint: disable=invalid-name, too-few-public-methods, missing-docstring
```
huzecong: I don't think this is necessary.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,477 @@
...
+import numpy as np
+import torch
+from torch import Tensor
+import torch.nn as nn
+from torch.optim.lr_scheduler import ExponentialLR
+
+import texar.torch as tx
+from texar.torch.modules import BasicRNNDecoderOutput, TransformerDecoderOutput
+from texar.torch.custom import MultivariateNormalDiag
+from texar.torch.modules.decoders.decoder_helpers import Helper
```
huzecong: We don't expose internals to users. I think these imports are all unnecessary, since they can be accessed via `tx.`.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,477 @@
...
+
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--config',
+                    type=str,
+                    default=None,
+                    help="The config to use.")
```
huzecong: Could you format these like you did for the GPT-2 example before? Make it more compact.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,477 @@
...
+            linear_layer_dim=self.encoder.cell.hidden_size * 2)
+
+        self.mlp_linear_layer = nn.Linear(
+            config_model.latent_dims,
+            sum_state_size)
+
+    def forward(self, data_batch: tx.data.Batch,
+                kl_weight: float, start_tokens: torch.LongTensor,
+                end_token: int) \
+            -> Dict:
```
huzecong: Don't wrap when it's unnecessary. Also, let's not use incomplete types. In this case, the return type could be `Dict[str, torch.Tensor]`.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,477 @@
...
+                max_decoding_length=max_decoding_length)
+        return outputs
+
+
+class VAEData(tx.data.MonoTextData):
+    r""" Wrap `tx.data.MonoTextData` to implement data preprocessing.
+    """
+    def process(self, raw_example: List[str]) -> List[str]:
+        # Truncates sentences and appends BOS/EOS tokens.
+        words = super().process(raw_example[1:-1])
```
huzecong: I think the extra blank token problem is already fixed in the master branch? Is this still necessary?

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,477 @@
...
+    """
+    config: Any = importlib.import_module(args.config)
+    device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+
+    train_data = VAEData(config.train_data_hparams,
+                                      device=device)
+    val_data = VAEData(config.val_data_hparams,
+                                    device=device)
+    test_data = VAEData(config.test_data_hparams,
+                                     device=device)
```
huzecong: No need to wrap here as well. If we were to wrap, this line should align with the opening bracket on the previous line. Try using auto-formatting tools in your IDE.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,477 @@
...
+
+        print('Output done')
+        fh.close()
+
+    if args.mode == "predict":
+        _generate(start_tokens, end_token, args.out)
+        return
+    # Counts trainable parameters
+    total_parameters = 0
+    for _, variable in model.named_parameters():
```
huzecong: If you don't need names, just call `model.parameters()`.

examples/vae_text/vae_train.py
```diff
@@ -0,0 +1,477 @@
...
+                scheduler.step()
+                print(f"-----\nchange lr, old lr: {old_lr}, "
+                      f"new lr: {new_lr}\n-----")
+
+                decay_cnt += 1
+                if decay_cnt == max_decay:
+                    break
+
+    print('\nbest testing nll: %.4f, best testing ppl %.4f\n' %
+        (best_nll, best_ppl))
```
huzecong: Let's also use f-strings for this one.

texar/torch/modules/connectors/connectors.py
```diff
@@ -144,19 +140,14 @@ def _mlp_transform(inputs: TensorStruct,
...
     if linear_layer is not None:
-        fc_output = linear_layer(fc_output)
+        fc_output = linear_layer(concat_input)
     if activation_fn is not None:
         fc_output = activation_fn(fc_output)
-
-    flat_output = torch.split(fc_output, size_list, dim=1)
+    elif linear_layer is None and activation_fn is None:
+        fc_output = concat_input
+    flat_output = split(fc_output, size_list, dim=1)
```
huzecong: Don't import individual functions from `torch`. Just call `torch.split`.

huzecong: And this part was changed by me, please revert it back. Your implementation here has problems: what if `linear_layer` is `None` but `activation_fn` is not?

texar/torch/modules/connectors/connectors.py
```diff
@@ -128,12 +128,8 @@ def _mlp_transform(inputs: TensorStruct,
     """
     # Flatten inputs
     flat_input = nest.flatten(inputs)
-    if len(flat_input[0].size()) == 1:
-        batch_size = 1
-    else:
-        batch_size = flat_input[0].size(0)
-    flat_input = [x.view(-1, x.size(-1)) for x in flat_input]
-    concat_input = torch.cat(flat_input, 0)
+    flat_input = [x.view(-1, np.prod(list(x.size())[1:])) for x in flat_input]
```
huzecong: I think the original implementation is correct here, since linear layers work on the final dimension.

examples/vae_text/config_lstm_ptb.py
TomNong: Got it, will fix.

examples/vae_text/vae_train.py
TomNong: Got it.

examples/vae_text/vae_train.py
TomNong: Sure, this will be fixed.

examples/vae_text/vae_train.py
TomNong: Got that.

examples/vae_text/vae_train.py
TomNong: for the `texar.torch.custom`, do we want to include it into modules?

examples/vae_text/vae_train.py
TomNong: Yes, it is redundant now.

examples/vae_text/vae_train.py
TomNong: Actually pylint used to report that, somehow it does not report it recently.

examples/vae_text/README.md
TomNong: Fixed that, thank!

texar/torch/modules/connectors/connectors.py
```diff
@@ -105,7 +105,6 @@ def _mlp_transform(inputs: TensorStruct,
                    ) -> Any:
     r"""Transforms inputs through a fully-connected layer that creates
     the output with specified size.
-
     Args:
```
huzecong: I think the blank line above is required.

TomNong:
@huzecong  thanks. One possible reason I think is that the checkpoint needs to store more information than current version, since we are using more than `learning_rate` to determine the stop of the training.

ZhitingHu:
Then why can't we store more info in order to recover the results? what's the difficulty here?

TomNong:
@ZhitingHu yeah... It is actually used to handle the case that training process pauses. I think we can add a parameter for continue training checkpoints, I'll make that.

ZhitingHu:
Did you mean the current code can reproduce the TF results if training is not paused in the middle?

TomNong:
@ZhitingHu Yes, actually "Yahoo-Lstm" once reached `ppl 68.95` in the middle, which is better then current `75.21`.

ZhitingHu:
Then report the results without interrupted training

TomNong:
Got it.

examples/vae_text/README.md
```diff
@@ -42,7 +42,7 @@ Here `--model` specifies the saved model checkpoint, which is saved in `./models
 
 |Dataset    |Metrics   | VAE-LSTM |VAE-Transformer |
 |---------------|-------------|----------------|------------------------|
-|Yahoo | Test PPL<br>Test NLL | 75.21<br>336.41 |67.81<br>328.34|
+|Yahoo | Test PPL<br>Test NLL | 69.42<br>338.65 |67.81<br>328.34|
```
huzecong: What about the Transformer results? Are those still running?

TomNong:
This PR is not yet ready for review.

TomNong:
This PR is ready for review now.

huzecong:
Results now look reasonable to me. Let's merge this.
After we merge this, can you take a look at the Executor module and think about how we can migrate this example to using Executors?

PR #137, opened 2019-08-02
Support RoBERTa in BERT modules

### LINKED ISSUE #131: Support RoBERTa in pretrained modules
https://github.com/pytorch/fairseq/tree/master/examples/roberta

gpengzhi:
The architecture of `RoBERTa` is very similar to that of `BERT`, except:
1. `RoBERTa` removes the next sentence prediction objective so that there is no `segment_embedding` and `pooler` in the pre-training procedure.
2. `BERT` uses a character-level BPE vocabulary of size 30K. `RoBERTa` uses a larger byte-level BPE vocabulary containing 50K sub-word units.

Resolve #131


texar/torch/modules/pretrained/pretrained_bert.py
```diff
@@ -70,59 +83,86 @@ def _transform_config(cls, cache_dir: str) -> Dict[str, Any]:
...
+                    num_blocks = config_ckpt['num_hidden_layers']
+                    num_heads = config_ckpt['num_attention_heads']
+                    dropout_rate = config_ckpt['attention_probs_dropout_prob']
+                    residual_dropout = config_ckpt['hidden_dropout_prob']
+                    intermediate_size = config_ckpt['intermediate_size']
+                    hidden_act = config_ckpt['hidden_act']
+
+            elif file.endswith('model.pt'):
+                config_path = os.path.join(root, file)
+                device = torch.device(
```
huzecong: I don't think we need to specify `device` here, since we don't need the parameters here. We can just map everything to `'cpu'`.

texar/torch/modules/pretrained/pretrained_bert.py
```diff
@@ -70,59 +83,86 @@ def _transform_config(cls, cache_dir: str) -> Dict[str, Any]:
...
+                    position_size = config_ckpt['max_position_embeddings']
+                    embedding_dropout = config_ckpt['hidden_dropout_prob']
+                    num_blocks = config_ckpt['num_hidden_layers']
+                    num_heads = config_ckpt['num_attention_heads']
+                    dropout_rate = config_ckpt['attention_probs_dropout_prob']
+                    residual_dropout = config_ckpt['hidden_dropout_prob']
+                    intermediate_size = config_ckpt['intermediate_size']
+                    hidden_act = config_ckpt['hidden_act']
+
+            elif file.endswith('model.pt'):
```
huzecong: This check looks a bit hacky. I think it might be better to refactor the interfaces and also pass in `pretrained_model_name` to `_transform_config` and `_init_from_checkpoint`.

texar/torch/modules/pretrained/pretrained_bert.py
```diff
@@ -231,3 +277,78 @@ def _init_from_checkpoint(self, cache_dir: str, **kwargs):
...
+            ],
+            'self_attn.in_proj_bias': [
+                'encoder.self_attns.{}.Q_dense.bias',
+                'encoder.self_attns.{}.K_dense.bias',
+                'encoder.self_attns.{}.V_dense.bias'
+            ],
+        }
+
+        checkpoint_path = os.path.abspath(os.path.join(cache_dir, 'model.pt'))
+        device = torch.device(
```
huzecong: This has potential issues because the model (`self`) might not be on the device as defined here. For robustness we can fetch a parameter from the model and use its parameter, probably like this:
```python
device = next(self.parameters()).device
```

ZhitingHu:
A quick question related to our last chat -- can RoBERTa support `segment_embedding` (even if model is pretrained w/o `segment_embedding`)? If not, can the model be applied to two-sentence input tasks (e.g., QA where the inputs include a passage and a question)? How?

gpengzhi:
@ZhitingHu In their paper, they showed the results for two-sentence input tasks (e.g. question answering). So RoBERTs can support sentence-pair tasks. To the best of my knowledge, they do not introduce the segment embedding in the fine-tuning procedure even for those sentence-pair tasks. 

In the data preparation step, they will add a separator token in between two sentences [here](https://github.com/pytorch/fairseq/blob/master/fairseq/tasks/sentence_prediction.py#L124). But the model is still fixed [here](https://github.com/pytorch/fairseq/blob/master/fairseq/models/roberta/model.py#L211). `num_segments` is always 0 which will remove `segment_embedding` in `Transformer` [here](https://github.com/pytorch/fairseq/blob/master/fairseq/modules/transformer_sentence_encoder.py#L111).

texar/torch/modules/classifiers/bert_classifier.py
```diff
@@ -47,7 +47,8 @@ class BERTClassifier(ClassifierBase):
             of a pre-trained model to load selected in the list of:
             `bert-base-uncased`, `bert-large-uncased`, `bert-base-cased`,
             `bert-large-cased`, `bert-base-multilingual-uncased`,
-            `bert-base-multilingual-cased`, `bert-base-chinese`.
+            `bert-base-multilingual-cased`, `bert-base-chinese`,
+            `roberta-base`, `roberta-large`, `roberta-large-mnli`.
```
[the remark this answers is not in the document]
ZhitingHu: Also, each category needs a simple, high-level description to help users quickly choose which one to use. For example, for `RoBERTa`, we can say sth like "This is a model with the largely same original BERT architecture but trained on more data  ...." (pls refer to their paper abstract), also mention that "Differing from the original BERT, RoBERTa does not have segmentation embedding". 

Please make all these organized in a structured way for easy-reading

texar/torch/modules/encoders/bert_encoders.py
```diff
@@ -41,15 +41,16 @@ class BERTEncoder(EncoderBase, PretrainedBERTMixin):
     :class:`~texar.torch.modules.encoders.TransformerEncoder` and a dense
     pooler.
 
-    This module supports the architecture first proposed
-    in `(Devlin et al.)` BERT.
+    This module supports the architectures first proposed
+    in `(Devlin et al.)` BERT and `(Liu et al.)` RoBERTa.
```
ZhitingHu: Since multiple places refer to these BERT variants, we can consider putting the structured note I mentioned above to one place, and point users to it. Is there sth like `BERTBase` class for us to put this "meta" info there?

texar/torch/modules/encoders/bert_encoders.py
```diff
@@ -41,15 +41,16 @@ class BERTEncoder(EncoderBase, PretrainedBERTMixin):
...
 
     Args:
         pretrained_model_name (optional): a str with the name
             of a pre-trained model to load selected in the list of:
             `bert-base-uncased`, `bert-large-uncased`, `bert-base-cased`,
             `bert-large-cased`, `bert-base-multilingual-uncased`,
-            `bert-base-multilingual-cased`, `bert-base-chinese`.
+            `bert-base-multilingual-cased`, `bert-base-chinese`,
+            `roberta-base`, `roberta-large`, `roberta-large-mnli`.
```
ZhitingHu: Make the list more structured, and refer users to the above structured note for more details.

.gitignore
```diff
@@ -222,7 +222,7 @@ docs/_build
 
 ### Project ###
 /data/
-/texar_download/
+texar_download/
```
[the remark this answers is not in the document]
gpengzhi: Since we added one more folder layer (`torch`) in `texar`, `texar_download` is currently in `texar` folder due to function `default_download_dir()` in `pretrained_base.py`. I wll update `default_download_dir()` accordingly.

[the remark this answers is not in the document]
huzecong: I see. Yes, that would be great.

texar/torch/modules/pretrained/pretrained_bert.py
```diff
@@ -76,43 +76,47 @@ class PretrainedBERTMixin(PretrainedMixin, ABC):
...
+                        embedding_dropout = config_ckpt['hidden_dropout_prob']
+                        num_blocks = config_ckpt['num_hidden_layers']
+                        num_heads = config_ckpt['num_attention_heads']
+                        dropout_rate = config_ckpt[
+                            'attention_probs_dropout_prob']
+                        residual_dropout = config_ckpt['hidden_dropout_prob']
+                        intermediate_size = config_ckpt['intermediate_size']
+                        hidden_act = config_ckpt['hidden_act']
+
+        if pretrained_model_name.startswith('roberta'):
```
huzecong: Let's change this to an `elif` to make the logic more clear.

texar/torch/modules/pretrained/pretrained_bert.py
```diff
@@ -173,10 +177,11 @@ def _transform_config(cls, cache_dir: str) -> Dict[str, Any]:
...
         return configs
 
-    def _init_from_checkpoint(self, cache_dir: str, **kwargs):
-        if cache_dir.split('/')[-1].startswith('bert'):
+    def _init_from_checkpoint(self, pretrained_model_name: str,
+                              cache_dir: str, **kwargs):
+        if pretrained_model_name.startswith('bert'):
             self._init_bert_from_checkpoint(cache_dir)
-        if cache_dir.split('/')[-1].startswith('roberta'):
+        if pretrained_model_name.startswith('roberta'):
```
huzecong: Same here.

ZhitingHu:
How is the PR? Would it be done any time soon?

gpengzhi:
Yes. I will update this PR soon.

PR #138, opened 2019-08-05
Enhancement of Convolution Networks

### LINKED ISSUE #136: About the Conv1DClassifier
Hello, I'm trying to reduplicate an example (which is originally implemented with `texar`) using `texar-pytorch`. But I got confused about the `Conv1DClassifier`.

1. The first question is about the [documents](https://texar-pytorch.readthedocs.io/en/latest/code/modules.html#texar.torch.modules.Conv1DNetwork):

image

However there doesn't seem to be a hyperparameter called `filter`. Maybe it means the `out_channels`? 

2. Currently the `Conv1DClassifier` requires an inputs with a shape of `[batch_size, channels, length]`, which means I have to manually transpose the input because in most case it have a shape like `[batch_size, length, dim]`.  I understand that the requirement resulting from pyTorch's `torch.nn.Conv1d`, but I wonder if the classifier can do the transposing job for me so that its  input style may keep in line with other classifiers (always set the `length` as the second dim). 

3. In the Tensorflow version of `texar`, I can set `'other_conv_kwargs': {'padding': 'same'}` to make sure the output shape will be identical to the input shape, but I can't do this in `texar-pytorch` since `torch.nn.Conv1d` only accept an integer for the padding numbers. Now I have 3 different kernel sizes [3, 4, 5] and my workaround is `'other_conv_kwargs': {'padding': 2}`. But I wonder if I can manually set different paddings for different kernel sizes like `padding: [1, 2, 2]`. It would be even better if `texar` can calculate the expected paddings for me.

4. the `Conv1DClassifier` will add a default hyperparameter:

https://github.com/asyml/texar-pytorch/blob/507932c899ca3a8663479b31efc3a41bc7180693/texar/torch/modules/classifiers/conv_classifiers.py#L136-L143

When I set `'num_dense_layers': 0`, The `Linear` layer below will still have a hyperparameter `in_features`, which is totally wrong because there isn't any dense layer in the `Conv1DEncoder`.

https://github.com/asyml/texar-pytorch/blob/507932c899ca3a8663479b31efc3a41bc7180693/texar/torch/modules/classifiers/conv_classifiers.py#L71-L85

In fact in this situation, the correct `in_features` should be the number of different kernel sizes times the number of out channels. Say, I have `'kernel_size': [3, 4, 5]` and `'out_channels': 128`, after Encoder-Flatten-Dropout, the `in_features` should be len([3,4,5]) * 128 = 384

AvinashBukkittu:
This PR 

- Corrects a few mistakes in the docs for Convolution Networks

- Adds `data_format` capability. User can now use `channels_first` and `channels_last` tensors. This option can be passed in forward function. If not passed, it will be picked from hyperparameters

- Handles the difference in `data_format` requirements between convolution network and `mask_sequences` [This was an existing issue which was unnoticed]

- Makes `other_conv_kwargs` and `other_pool_kwargs` accept list as well as dict. If dict, the same property is applied to all layers. If list, then individual layers get their own `kwargs`

- Fixes a bug - When `num_dense_layers < 0`, `in_features` of logits layer equals the `out_features` of `Flatten` layer. Adds a test case for this.

This PR closes #136

### DIFF
--- a/texar/torch/modules/classifiers/conv_classifiers.py
+++ b/texar/torch/modules/classifiers/conv_classifiers.py
@@ -175,8 +175,10 @@ class Conv1DClassifier(ClassifierBase):
                 from inputs automatically.
             data_format (optional): Data type of the input tensor. If
                 ``channels_last``, the last dimension will be treated as channel
-                dimension. If ``channels_first``, first dimension will be
-                treated as channel dimension. Defaults to None.
+                dimension so the size of the :attr:`input` should be
+                `[batch_size, X, channel]`. If ``channels_first``, first
+                dimension will be treated as channel dimension so the size
+                should be `[batch_size, channel, X]`. Defaults to None.
                 If None, the value will be picked from hyperparameters.
 
         Returns:
--- a/texar/torch/modules/classifiers/conv_classifiers.py
+++ b/texar/torch/modules/classifiers/conv_classifiers.py
@@ -69,9 +69,12 @@ class Conv1DClassifier(ClassifierBase):
         self._num_classes = self._hparams.num_classes
         if self._num_classes > 0:
             if self._hparams.num_dense_layers <= 0:
+                if in_features is None:
+                    raise ValueError("'in_features' is required for logits "
+                                     "layer when 'num_dense_layers' <= 0")
                 self._encoder.append_layer({"type": "Flatten"})
                 ones = torch.ones(1, in_channels, in_features)
-                input_size = self._encoder._infer_dense_layer_input_size(ones)
+                input_size = self._encoder._infer_dense_layer_input_size(ones)  # pylint: disable=protected-access
                 self.hparams.logit_layer_kwargs.in_features = input_size[1]
 
             logit_kwargs = self._hparams.logit_layer_kwargs
@@ -170,11 +173,11 @@ class Conv1DClassifier(ClassifierBase):
                 first be masked out before feeding to the layers.
             dtype (optional): Type of the inputs. If not provided, infers
                 from inputs automatically.
-            data_format(optional): Data type of the input tensor. If
+            data_format (optional): Data type of the input tensor. If
                 ``channels_last``, the last dimension will be treated as channel
-                 dimension. If ``channels_first``, first dimension will be
-                 treated as channel dimension. Defaults to None.
-                 If None, the value will be picked from hparams.
+                dimension. If ``channels_first``, first dimension will be
+                treated as channel dimension. Defaults to None.
+                If None, the value will be picked from hyperparameters.
 
         Returns:
             A tuple ``(logits, pred)``, where
--- a/texar/torch/modules/classifiers/conv_classifiers.py
+++ b/texar/torch/modules/classifiers/conv_classifiers.py
@@ -70,6 +70,9 @@ class Conv1DClassifier(ClassifierBase):
         if self._num_classes > 0:
             if self._hparams.num_dense_layers <= 0:
                 self._encoder.append_layer({"type": "Flatten"})
+                ones = torch.ones(1, in_channels, in_features)
+                input_size = self._encoder._infer_dense_layer_input_size(ones)
+                self.hparams.logit_layer_kwargs.in_features = input_size[1]
 
             logit_kwargs = self._hparams.logit_layer_kwargs
             if logit_kwargs is None:
@@ -146,7 +149,8 @@ class Conv1DClassifier(ClassifierBase):
     def forward(self,  # type:ignore
                 input: torch.Tensor,
                 sequence_length: Union[torch.LongTensor, List[int]] = None,
-                dtype: Optional[torch.dtype] = None) \
+                dtype: Optional[torch.dtype] = None,
+                data_format: Optional[str] = None) \
             -> Tuple[torch.Tensor, torch.Tensor]:
         r"""Feeds the inputs through the network and makes classification.
 
@@ -166,6 +170,11 @@ class Conv1DClassifier(ClassifierBase):
                 first be masked out before feeding to the layers.
             dtype (optional): Type of the inputs. If not provided, infers
                 from inputs automatically.
+            data_format(optional): Data type of the input tensor. If
+                ``channels_last``, the last dimension will be treated as channel
+                 dimension. If ``channels_first``, first dimension will be
+                 treated as channel dimension. Defaults to None.
+                 If None, the value will be picked from hparams.
 
         Returns:
             A tuple ``(logits, pred)``, where
@@ -180,7 +189,7 @@ class Conv1DClassifier(ClassifierBase):
               ``{0, 1}``.
         """
         logits = self._encoder(input, sequence_length=sequence_length,
-                               dtype=dtype)
+                               dtype=dtype, data_format=data_format)
 
         num_classes = self._hparams.num_classes
         is_binary = num_classes == 1
--- a/texar/torch/modules/networks/conv_networks.py
+++ b/texar/torch/modules/networks/conv_networks.py
@@ -583,17 +583,33 @@ class Conv1DNetwork(FeedForwardNetworkBase):
         if data_format is None:
             data_format = self.hparams["data_format"]
 
-        if data_format == "channels_last":
+        if data_format == "channels_first":
+            # masking requires channels in last dimension
             input = input.permute(0, 2, 1)
 
-        if sequence_length is not None:
-            input = mask_sequences(input, sequence_length,
-                                   dtype=dtype, time_major=False)
-        output = super().forward(input)
+            if sequence_length is not None:
+                input = mask_sequences(input, sequence_length,
+                                       dtype=dtype, time_major=False)
 
-        if (self._hparams.num_dense_layers <= 0
-                and data_format == "channels_last"):
-            output = output.permute(0, 2, 1)
+            # network is constructed for channel first tensors
+            input = input.permute(0, 2, 1)
+
+            output = super().forward(input)
+
+        elif data_format == "channels_last":
+            if sequence_length is not None:
+                input = mask_sequences(input, sequence_length,
+                                       dtype=dtype, time_major=False)
+
+            input = input.permute(0, 2, 1)
+
+            output = super().forward(input)
+
+            # transpose only when no dense layers
+            if self._hparams.num_dense_layers <= 0:
+                output = output.permute(0, 2, 1)
+        else:
+            raise ValueError("Invalid 'data_format'")
 
         return output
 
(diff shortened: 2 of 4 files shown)


texar/torch/modules/networks/conv_networks.py
```diff
@@ -517,13 +572,46 @@ def forward(self,  # type: ignore
...
+                input = mask_sequences(input, sequence_length,
+                                       dtype=dtype, time_major=False)
+
+            input = input.permute(0, 2, 1)
+
+            output = super().forward(input)
+
+            # transpose only when no dense layers
+            if self._hparams.num_dense_layers <= 0:
+                output = output.permute(0, 2, 1)
```
huhk-sysu: What if `Conv1DClassifier` append a `Flatten` layer and a `Linear` layer for the encoder? In this case the `output` may have a shape [batch_size, ...] and can't be permuted.

AvinashBukkittu: No problem. We value the feedback. Good find! A more elegant check I think is `if output.dim() == 3`

texar/torch/modules/classifiers/conv_classifiers.py
```diff
@@ -146,7 +152,8 @@ def default_hparams() -> Dict[str, Any]:
     def forward(self,  # type:ignore
                 input: torch.Tensor,
                 sequence_length: Union[torch.LongTensor, List[int]] = None,
-                dtype: Optional[torch.dtype] = None) \
+                dtype: Optional[torch.dtype] = None,
+                data_format: Optional[str] = None) \
```
ZhitingHu: what's the conventional ordering of `data_format` and `dtype`? Could you check pytorch/TF APIs?

texar/torch/modules/classifiers/conv_classifiers.py
```diff
@@ -166,6 +173,11 @@ def forward(self,  # type:ignore
                 first be masked out before feeding to the layers.
             dtype (optional): Type of the inputs. If not provided, infers
                 from inputs automatically.
+            data_format (optional): Data type of the input tensor. If
+                ``channels_last``, the last dimension will be treated as channel
+                dimension. If ``channels_first``, first dimension will be
+                treated as channel dimension. Defaults to None.
+                If None, the value will be picked from hyperparameters.
```
ZhitingHu: Make the input size clearer, e..g, 

If ``channels_last``, the last dimension will be treated as channel dimension, so the size of :attr:`input` should be `[batch_size, X, channel]` , ...

texar/torch/modules/networks/conv_networks.py
```diff
@@ -172,10 +180,29 @@ def default_hparams():
...
+                 instead a dict, then the same dict gets applied to all the
+                 convolution layers at index i.
+
+           `"data_format"`: str, optional
+               Data format of the input tensor. Defaults to ``channels_first``
+               denoting the first dimension to be the channel dimension. Set it
+               to ``channels_last`` to treat last dimension as the channel
+               dimension. This argument can also be passed in ``forward``
+               function, in which case the value specified here will be
+               ignored.
```
ZhitingHu: Same here, make the expected input size clear.

texar/torch/modules/networks/conv_networks.py
```diff
@@ -197,9 +224,15 @@ def default_hparams():
                layers will have the same stride. If a list, the list length
                must equal ``"num_conv_layers"``.
 
-           `"other_pool_kwargs"`: dict, optional
+           `"other_pool_kwargs"`: list or dict, optional
                Other keyword arguments for pooling layer class constructor.
 
+               - If a dict, the same dict is applied to all the pooling layers.
+
+               - If a list, the length must equal ``"num_conv_layers"``. The
```
ZhitingHu: Can this (or is it necessary to) be a nested list as w/ `other_conv_kwargs`?

texar/torch/modules/networks/conv_networks.py
```diff
@@ -349,11 +390,17 @@ def _build_conv1d_hparams(self, in_channels, pool_hparams):
...
-        other_kwargs = self._hparams.other_conv_kwargs or {}
+        other_kwargs = self._hparams.other_conv_kwargs
         if isinstance(other_kwargs, HParams):
             other_kwargs = other_kwargs.todict()
-        if not isinstance(other_kwargs, dict):
-            raise ValueError("hparams['other_conv_kwargs'] must be a dict.")
+            other_kwargs = _to_list(other_kwargs, "other_conv_kwargs", nconv)
+        elif isinstance(other_kwargs, (list, tuple)):
+            if len(other_kwargs) != nconv:
+                raise ValueError("len(hparams['other_conv_kwargs']) must be "
```
ZhitingHu: `len(hparams['other_conv_kwargs'])` looks a bit complicated. Use sth like: The length of the `hparams['other_conv_kwargs']` list

ZhitingHu:
Is the new issue about `time_major=False` included in this PR?

AvinashBukkittu:
> Is the new issue about `time_major=False` included in this PR?

Yes, it is included. I have updated the description of this PR to mention this issue.

texar/torch/modules/networks/conv_networks.py
huzecong: I think a more appropriate check would be `if output.dim() > 3`. Also, this can be rewritten as `output.transpose(1, 2)` which works for more than 3 dimensions.

texar/torch/modules/classifiers/conv_classifiers.py
huzecong: Conventionally there's no `data_format` argument in PyTorch. There is, however, `layout` which is positioned after `dtype`. I actually think it's best if we make these arguments keyword-only, but for now I think we can keep it as is for consistency.

texar/torch/modules/classifiers/conv_classifiers.py
```diff
@@ -69,7 +69,13 @@ def __init__(self, in_channels: int, in_features: Optional[int] = None,
         self._num_classes = self._hparams.num_classes
         if self._num_classes > 0:
             if self._hparams.num_dense_layers <= 0:
+                if in_features is None:
+                    raise ValueError("'in_features' is required for logits "
+                                     "layer when 'num_dense_layers' <= 0")
                 self._encoder.append_layer({"type": "Flatten"})
+                ones = torch.ones(1, in_channels, in_features)
+                input_size = self._encoder._infer_dense_layer_input_size(ones)  # pylint: disable=protected-access
```
huzecong: Now that we have `output_size`, do we still need this?

texar/torch/modules/networks/conv_networks.py
AvinashBukkittu: Currently, this cannot be a nested list. The same holds true for `pool_size` and `pool_stride`

texar/torch/modules/networks/conv_networks.py
AvinashBukkittu: I think you meant `output.dim() >= 3`? However, since we use `Conv1D`, our tensors are 3D, we cannot perform Conv1D on higher than 3D tensors, right?

texar/torch/modules/classifiers/conv_classifiers.py
AvinashBukkittu: I also quickly checked Tensorflow APIs. I couldn't find functions which take `data_format` and `dtype` as arguments.

texar/torch/modules/classifiers/conv_classifiers.py
AvinashBukkittu: I am not sure I understand the logic of `output_size` method. I think this was added after the PR for Convolution Networks was merged. It returns `self._hparams.out_channels * len(self._hparams.kernel_size)` which is wrong. First `self._hparams.out_channels` can be a list. Even if it is not a list, when we have more than one convolution layers (For e.g., a merge layer followed by a single convolution layer), the above returned value is wrong. One way to know the output size is to pass the dummy tensor through the network, another way is to work it out through a mathematical formula.

texar/torch/modules/classifiers/conv_classifiers.py
AvinashBukkittu: Updated in d2c2634

texar/torch/modules/networks/conv_networks.py
AvinashBukkittu: Updated in d2c2634

texar/torch/modules/networks/conv_networks.py
AvinashBukkittu: Updated in d2c2634

texar/torch/modules/classifiers/conv_classifiers.py
huzecong: I see. Could you try to correct that implementation? I would prefer we work out a formula to compute the size, rather than having to do a forward pass.

texar/torch/modules/networks/conv_networks.py
huzecong: Ah yes you're right. Sorry. Then let's add a check to ensure that the input does not contain more than 3 dimensions.

texar/torch/modules/classifiers/conv_classifiers.py
AvinashBukkittu: The formula can be a bit tricky to write, as it has to consider several cases of MergeLayer combinations, stride sizes, kernel sizes of several layers. Can we do it as part of a separate PR? A dummy pass is a simple solution. And regarding the output size, we actually cannot know it unless we know the size of input tensor. That is the reason we take `in_channels` and `in_features` as part of arguments in the constructor.

texar/torch/modules/networks/conv_networks.py
AvinashBukkittu: Added a check in 373773c

texar/torch/modules/classifiers/conv_classifiers.py
huzecong: We could do it as a separate PR, but I think it's something we should have. 
If there are no dense layers then I think the output size should just be the output channel count (at least in NLP this makes sense), and we should be able to know that given `in_channels` alone (there's actually a formula on PyTorch docs).
If dense layers exist, we'll just use the size of the output of the dense layer.

AvinashBukkittu: Added a fix in fb85495

texar/torch/modules/classifiers/conv_classifiers.py
ZhitingHu: what does make these arguments keyword-only" mean?

huzecong: It means that to specify these arguments, one must write the argument name in full:

```python
def func(input, *, dtype=None):
	pass

func(tensor)                     # Correct; dtype is None
func(tensor, torch.float)        # Wrong
func(tensor, dtype=torch.float)  # Correct
func(tensor, **{"dtype": torch.float})  # Also correct
```

Also see: https://python-3-for-scientists.readthedocs.io/en/latest/python3_advanced.html

PR #141, opened 2019-08-05
Speed up CI

AvinashBukkittu:
This PR is an attempt to speed up CI

- Removes TF dependency

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ python:
 
 install:
   - pip install --upgrade pip
-  - pip install --progress-bar off .[torch,extras] tensorflow
+  - pip install --progress-bar off .[torch,extras]
   - pip install pylint==2.3.1 flake8==3.7.7
   - pip install mypy==0.720
   - pip install pytest


gpengzhi:
@AvinashBukkittu Thanks for the fix! Do we plan to merge it?

AvinashBukkittu:
> @AvinashBukkittu Thanks for the fix! Do we plan to merge it?

Not yet. We can try out a few things to see how we can reduce the running time. Few of the possible things we can try are

- Try to parallelize `pytest`
- Instead of installing the libraries every time, I was wondering if we can dockerize it and cache the docker image or something
- Try splitting the jobs into smaller jobs

gpengzhi:
Sounds great. Maybe you could set up CI in your forked repo so that you can try things there at first.

huzecong:
Can we close this in favor of #148?

AvinashBukkittu:
> Can we close this in favor of #148?

Sure, I will close it for now. But I think even without tensorflow it takes 10min. I will try to make it faster in the future. But it is of low priority right now.

PR #143, opened 2019-08-06
Update Connector

TomNong:
* Update implement of mlp transform.
* Update docstring.


texar/torch/modules/connectors/connectors.py
```diff
@@ -128,12 +124,8 @@ def _mlp_transform(inputs: TensorStruct,
     """
     # Flatten inputs
     flat_input = nest.flatten(inputs)
-    if len(flat_input[0].size()) == 1:
-        batch_size = 1
-    else:
-        batch_size = flat_input[0].size(0)
-    flat_input = [x.view(-1, x.size(-1)) for x in flat_input]
-    concat_input = torch.cat(flat_input, 0)
+    flat_input = [x.view(-1, np.prod(list(x.size())[1:])) for x in flat_input]
```
huzecong: I believe I've said this before. This can be written as:
```python
x.view(x.size(0), -1)
```

texar/torch/modules/connectors/connectors.py
```diff
@@ -167,6 +152,18 @@ def _mlp_transform(inputs: TensorStruct,
     return output
 
 
+def _prod(x: Tuple[int, ...]) -> int:
```
ZhitingHu: Why do we need to define such a basic operator? No numpy API  out there?

texar/torch/modules/connectors/connectors.py
```diff
@@ -76,7 +76,7 @@ def _get_tensor_depth(x: torch.Tensor) -> int:
     Args:
         x: A tensor.
     """
-    return int(np.prod(x.size()[1:]))
+    return np.prod(x.size()[1:])
```
huzecong: This function isn't being used at all. Do we still need this?

texar/torch/modules/connectors/connectors.py
```diff
@@ -16,8 +16,8 @@
 """
 
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
-
```
huzecong: As a reference, the preferred way to format imports is:

- Split imports into three parts: builtin packages, third-party packages, our own packages. Parts are listed in this order and different parts are separated by one empty line.
- Within each part, sort imports in lexicographical order by their package names. For the same library `xxx`, `import xxx` should come before `from xxx import yyy`.
- My personal preference is to also sort imported identifiers in `from ...` in case-sensitive lexicographical order. But you don't have to do this.

Before this change, the imports in this file were correct.

texar/torch/modules/connectors/connectors.py
```diff
@@ -516,7 +501,7 @@ class ReparameterizedStochasticConnector(ConnectorBase):
 
     def __init__(self,
                  output_size: OutputSize,
-                 mlp_input_size: Union[torch.Size, MaybeTuple[int], int],
+                 mlp_input_size: MaybeTuple[int],
```
huzecong: Do we still accept `torch.Size`?

texar/torch/modules/connectors/connectors.py
TomNong: Great explanations, thanks!

TomNong:
Sorry I think I should not rebase on the master at this point, I will solve these commits.

texar/torch/modules/connectors/connectors.py
TomNong: Yeah, `torch.Size` can pass `Tuple` too, so I did this change, I re-think about it that specifying it makes more sense.

huzecong:
When will the PR be ready for review?

TomNong:
@huzecong Sorry I'll open a new one.

PR #153, opened 2019-08-13
Resolve #147 Fix doc outputsize

### LINKED ISSUE #147: Write clear docs for output_size methods. In few cases, we return the size of only one of the tensors.
Then the docstring "The final dimension(s) of :meth:forward output tensor(s)" should be updated. It's not exactly the size(s) of forward outputs, but instead more of case-by-case? E.g., sometimes it's the dimensions of some of the output tensors. 

And, what does "final dimension(s)" mean? 

Since it's case-by-case, all `output_size` should explicitly and clearly explain what it is in the particular classes.

_Originally posted by @ZhitingHu in https://github.com/asyml/texar/pull/182_

TomNong:
* Update docs.
* Update output_size func in bert classifer and gpt-2 classifier.


TomNong:
One thing I'm concerning is that we did omit the `output_size` doc string for some modules, let me know if we need to restore that, thanks!

texar/torch/modules/connectors/connector_base.py
```diff
@@ -63,4 +63,6 @@ def default_hparams():
 
     @property
     def output_size(self):
+        r"""The output size of :attr:`Connectors`.
```
huzecong: I don't think this one's required. Plus, there's no such thing as ``":attr:`Connectors`"``, it's either `"connectors"` or an explicit class ``":class:`SomeConnector`"``.

texar/torch/modules/embedders/embedder_base.py
```diff
@@ -142,9 +142,8 @@ def forward(self,  # type: ignore
...
     def output_size(self) -> int:
-        r"""The final dimension(s) of :meth:`forward` output tensor(s).
-
-        Here final dimension is ``1``, because here output tensor size equals
-        to the input tensor size.
+        r"""The feature size of :meth:`forward` output.
+        Equals to ``1`` because here output size equals
+        to the input size.
         """
         return 1
```
huzecong: I'm actually thinking whether we need `output_size` for this class. We use a size of 1 for classifiers because when `num_classes` is one we can interpret the output as having a size of `[batch_size, 1]` (although we squeeze that final dimension), but it's not the case here and having 1 does not really make sense.
Since no one will probably use this method, I'm thinking maybe we could just throw a `ValueError` for the modules where we can't calculate the output size.

texar/torch/modules/embedders/embedders.py
```diff
@@ -258,6 +258,9 @@ def num_embeddings(self) -> int:
 
     @property
     def output_size(self) -> int:
+        r"""The feature size of :meth:`forward` output :attr:`outputs`.
+        Same as embedding dimension's last element or itself.
```
huzecong: I don't understand this sentence. Could you rephrase it? And same for the position embedders.

huzecong: We might not need to mention `outputs` here because the `forward` method returns only one value.

texar/torch/modules/encoders/gpt2_encoder.py
```diff
@@ -293,4 +293,7 @@ def forward(self,  # type: ignore
 
     @property
     def output_size(self):
+        r"""The feature size of :meth:`forward` output
+        :attr:`output`.
```
huzecong: Same here.

texar/torch/modules/encoders/multihead_attention.py
```diff
@@ -255,4 +255,7 @@ def _combine_heads(self, x: torch.Tensor) -> torch.Tensor:
 
     @property
     def output_size(self):
+        r"""The feature size of :meth:`forward` output
+        :attr:`outputs`.
```
huzecong: And here.

texar/torch/modules/encoders/rnn_encoders.py
```diff
@@ -459,15 +459,19 @@ def output_layer(self) -> Optional[nn.Module]:
...
+        r"""The feature size of :meth:`forward` output :attr:`outputs`.
+        If output layer does not exist, equals to
+        :attr:`encoder.cell.hidden_size`, otherwise equals to
+        last dimension value of output layer output size.
+        """
         # TODO: We will change the implementation to
         # something that does not require a forward pass.
 
         dim = self._cell.hidden_size
         dummy_tensor = torch.Tensor(dim)
```
huzecong: Let's move the dummy tensor construction inside the `if`-branch. Same for the bi-RNNs.

texar/torch/modules/encoders/rnn_encoders.py
```diff
@@ -459,15 +459,19 @@ def output_layer(self) -> Optional[nn.Module]:
 
     @property
     def output_size(self) -> int:
+        r"""The feature size of :meth:`forward` output :attr:`outputs`.
+        If output layer does not exist, equals to
```
huzecong: You need a subject for this sentence.

texar/torch/modules/encoders/xlnet_encoder.py
```diff
@@ -269,6 +269,9 @@ def param_groups(self,
 
     @property
     def output_size(self):
+        r"""The feature size of :meth:`_forward` result
```
huzecong: Let's not mention `_forward` as we don't expect users to use that method. And let's be consistent with the docstring, in most of then you say "output", but here you say "result".

texar/torch/modules/networks/conv_networks.py
```diff
@@ -628,6 +628,8 @@ def _infer_dense_layer_input_size(self, input: torch.Tensor) -> torch.Size:
 
     @property
     def output_size(self) -> int:
+        r"""The feature size of :meth:`forward` output :attr:`output`.
```
huzecong: ``:attr:`output` `` might be unnecessary here too.

texar/torch/modules/networks/networks.py
```diff
@@ -87,12 +87,9 @@ def default_hparams():
...
-        r"""The final dimension(s) of :meth:`forward` output tensor(s).
-
-        Here final dimension equals to ``1`` if unable to get output size
-        from network layers.
+        r"""The feature size of network layers result.
+        If unable to get output size from network layers, equals to ``1``.
         """
-
         for layer in reversed(self._layers):
             size = get_output_size(layer)
```
huzecong: I'm not sure if this implementation is correct though. You could have arbitrary layers here, and receiving a size of `None` doesn't mean the layer output size is the same as its input size -- it just means we don't know how to compute it.

A better approach would be to make `get_output_size` work for more cases. For instance, we can also create a list of modules that does not alter the tensor size, among them activations and dropouts. We also need to check if the layer has `output_size` as a property, and if it does, return that property instead.

texar/torch/modules/regressors/xlnet_regressor.py
```diff
@@ -262,4 +262,7 @@ def forward(self,  # type: ignore
 
     @property
     def output_size(self) -> int:
+        r"""The feature size of :meth:`forward` output :attr:`preds`.
+        Equals to ``1`` because it is only determined by inputs.
```
huzecong: Let's use the same wording for the classifiers.

texar/torch/utils/utils.py
```diff
@@ -645,7 +645,7 @@ def get_instance_kwargs(kwargs: Kwargs, hparams: ParamDict) -> Kwargs:
 
 
 def get_output_size(instance_: nn.Module) -> Optional[int]:
-    r"""Return the final dimension of output size from :attr:`instance_`.
+    r"""Return the final dimension value of output size from :attr:`instance_`.
```
huzecong: What is a "dimension value of size"? Please be more clear.

texar/torch/modules/embedders/embedder_base.py
TomNong: Thanks, would you think we remain the function but throw out `ValueError`? In that case how about we use `AttributeError`?

huzecong: It depends on how we want to interpret this error:

- If the error occurs because of certain configurations or values, it's a `ValueError`.
- If the error occurs because the class in question does not support such operations, it's a `TypeError`.
- If the error occurs during an attribute lookup and fails because the instance in question does not have such attribute, it's an `AttributeError`.

In our case, using `AttributeError` might confuse the user. I think it should be either `ValueError` or `TypeError`, maybe `TypeError` is more appropriate here.

texar/torch/modules/networks/networks.py
TomNong: I see, how should we use the list of modules that does not alter the tensor size in computing `output_size`? Thanks!

TomNong: I may see what you mean, thanks!

texar/torch/modules/embedders/embedder_base.py
```diff
@@ -141,9 +141,5 @@ def forward(self,  # type: ignore
...
 
     @property
-    def output_size(self) -> int:
-        r"""The feature size of :meth:`forward` output.
-        Equals to ``1`` because here output size equals
-        to the input size.
-        """
-        return 1
+    def output_size(self):
+        raise ValueError("'output_size' can not be calculated.")
```
huzecong: Let's also give an explanation, saying that `output_size` cannot be calculated because it is equal to the input size.

texar/torch/modules/embedders/embedders.py
```diff
@@ -258,8 +258,10 @@ def num_embeddings(self) -> int:
 
     @property
     def output_size(self) -> int:
-        r"""The feature size of :meth:`forward` output :attr:`outputs`.
-        Same as embedding dimension's last element or itself.
+        r"""The feature size of :meth:`forward` output. If embedding
+        dimension is a ``list`` or ``tuple``, the feature size is
+        same as its last element. Otherwise if embedding dimension
+        is a ``int``, the feature size is same as itself.
```
huzecong:
```
The feature size of :meth:`forward` output. If the :attr:`dim`
hyperparameter is a ``list`` or ``tuple``, the feature size
equals its final dimension; otherwise, if :attr:`dim` is an
``int``, the feature size equals :attr:`dim`.
```

texar/torch/modules/encoders/rnn_encoders.py
```diff
@@ -460,16 +460,16 @@ def output_layer(self) -> Optional[nn.Module]:
     @property
     def output_size(self) -> int:
         r"""The feature size of :meth:`forward` output :attr:`outputs`.
-        If output layer does not exist, equals to
-        :attr:`encoder.cell.hidden_size`, otherwise equals to
-        last dimension value of output layer output size.
+        If output layer does not exist, the feature size equals to
+        :attr:`encoder.cell.hidden_size`, otherwise the feature size
+        equals to last dimension value of output layer output size.
```
huzecong: It's either "A equals B" or "A is equal to B".

texar/torch/modules/networks/networks.py
```diff
@@ -87,11 +87,20 @@ def default_hparams():
...
+                size = size_ext
+            if size > 1:
                 return size
-        return 1
+            elif size == 1:
+                ones_cnt += 1
+        if ones_cnt == len(self._layers):
+            return 1
+        else:
+            raise ValueError("'output_size' can not be calculated.")
```
huzecong: Have you tested this implementation? It doesn't look correct to me.

texar/torch/utils/utils.py
```diff
@@ -644,24 +650,29 @@ def get_instance_kwargs(kwargs: Kwargs, hparams: ParamDict) -> Kwargs:
...
-        if isinstance(instance_, t):
-            return l(instance_)
-    return None
+    for t, l in Type_size_lambda_map.items():
+        if isinstance(input_instance, t):
+            return l(input_instance)
+    for t in Type_size_keeper:
+        if isinstance(input_instance, t):
+            return 1
+    return -1
```
huzecong: What happened to returning `None`?

texar/torch/modules/networks/networks.py
TomNong: Would you have any suggestion?

texar/torch/modules/networks/networks.py
```diff
@@ -88,19 +88,19 @@ def default_hparams():
...
-        for layer in reversed(self._layers):
+        for i, layer in enumerate(reversed(self._layers)):
             size = get_output_size(layer)
             size_ext = getattr(layer, 'output_size', None)
             if size_ext is not None:
                 size = size_ext
-            if size > 1:
+            if size is None:
+                break
+            elif size > 1:
```
huzecong: This should be `> 0`.

texar/torch/modules/networks/networks.py
```diff
@@ -88,19 +88,19 @@ def default_hparams():
...
-                ones_cnt += 1
-        if ones_cnt == len(self._layers):
-            return 1
-        else:
-            raise ValueError("'output_size' can not be calculated.")
+            elif i == len(self._layers) - 1:
+                return -1
+
+        raise ValueError("'output_size' can not be calculated because "
+                         "it is equal to the input size.")
```
huzecong: The message is wrong. The reason is not that output size is equal to input size, but rather the module contains submodules whose output size we cannot determine.

Issue #146, opened 2019-08-08
Move the default download directory outside the package directory


gpengzhi:
The default download directory `texar_download` is currently under the package directory `texar-pytorch`. The preferred installation method now is `pip install .` instead of `pip install -e .`. We should store those large checkpoint files outside the package directory. We could probably set the default directory to `~/texar_download` (a folder under the home directory).

PR #159, opened 2019-08-18
Resolve #157 Doc cross references not work properly

### LINKED ISSUE #157: Doc cross references not work properly
E.g., [here](https://texar-pytorch.readthedocs.io/en/latest/code/modules.html#texar.torch.modules.BERTClassifier), "Please refer to PretrainedBERTMixin for all supported models" -- `PretrainedBERTMixin` does not correctly link to the PretrainedBERTMixin class. Other cross-refs (at least in pretrained modules) have the same issue

gpengzhi:
Resolve #157

### DIFF
--- a/texar/torch/modules/connectors/connectors.py
+++ b/texar/torch/modules/connectors/connectors.py
@@ -302,7 +302,7 @@ class ForwardConnector(ConnectorBase):
     :attr:`output_size`, or must have the same number of elements and be
     re-packable into the structure of :attr:`output_size`. Note that if input
     is or contains a ``dict`` instance, the keys will be sorted to pack in
-    deterministic order (See :func:`~texar.torch.utils.nest.pack_sequence_as`
+    deterministic order (See :func:`texar.torch.utils.nest.pack_sequence_as`
     for more details).
     """
 
--- a/texar/torch/modules/decoders/rnn_decoders.py
+++ b/texar/torch/modules/decoders/rnn_decoders.py
@@ -121,8 +121,8 @@ class BasicRNNDecoder(RNNDecoderBase[HiddenState, BasicRNNDecoderOutput]):
                 specified, you must subclass :class:`BasicRNNDecoder` and
                 override :meth:`embed_tokens`.
         cell (RNNCellBase, optional): An instance of
-            :class:`~texar.torch.core.RNNCellBase`. If `None` (default), a cell
-            is created as specified in :attr:`hparams`.
+            :class:`~texar.torch.core.cell_wrappers.RNNCellBase`. If `None`
+            (default), a cell is created as specified in :attr:`hparams`.
         output_layer (optional): An instance of :torch_nn:`Module`. Apply to
             the RNN cell output to get logits. If `None`, a :torch_nn:`Linear`
             layer is used with output dimension set to :attr:`vocab_size`.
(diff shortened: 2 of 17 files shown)


gpengzhi:
> Why did we remove the `~` prefix for certain references?

Since those references are not included in our `docs`, I think we should provide a full path for those references.

huzecong:
I see, that makes sense.

ZhitingHu:
Why ain't they not included in docs? It does not make sense to ask users to look into the code (or docstring in the code files). A cross-reference should not point to sth that does not exist in docs.

gpengzhi:
> Why ain't they included in docs? It does not make sense to ask users to look into the code (or docstring in the code files). A cross-reference should not point to sth that does not exist in docs.

If they are only used internally, why should we add them to `docs`?

ZhitingHu:
Then public docs should not refer to them --- they are just not for users to read.

ZhitingHu:
Of course, if the internal modules are referred to only by internal docs, then it's fine. If it's the case, it's good to merge the PR

gpengzhi:
> Then public docs should not refer to them --- they are just not for users to read. Users are not expected to know the existence of internal-only classes

Good point. I deleted or modified those internal-only classes and functions in docs.

texar/torch/modules/decoders/rnn_decoders.py
```diff
@@ -468,10 +468,9 @@ def default_hparams():
 
             `"type"`: str or class or instance
                 The attention type. Can be an attention class, its name or
-                module path, or a class instance. The class must be a subclass
-                of :class:`~texar.torch.core.AttentionMechanism`. If class name
-                is given, the class must be from modules :mod:`texar.torch.core`
-                or :mod:`texar.torch.custom`.
+                module path, or a class instance. If class name is given, the
+                class must be from modules :mod:`texar.torch.core` or
+                :mod:`texar.torch.custom`.
```
ZhitingHu: "The class must be a subclass of :class:`~texar.torch.core.AttentionMechanism`." --- This sentence is deleted? Is it cuz the class does not need to be the subclass of `AttentionMechanism`? If it does need, then we shall explicit say it, and expose `AttentionMechanism` to users, and probably say that common attention classes are already implemented, such as `BahdanauAttention` (refer users to the list: https://texar-pytorch.readthedocs.io/en/latest/code/core.html#attention-mechanism)

texar/torch/modules/connectors/connectors.py
```diff
@@ -302,8 +302,7 @@ class ForwardConnector(ConnectorBase):
     :attr:`output_size`, or must have the same number of elements and be
     re-packable into the structure of :attr:`output_size`. Note that if input
     is or contains a ``dict`` instance, the keys will be sorted to pack in
-    deterministic order (See :func:`~texar.torch.utils.nest.pack_sequence_as`
-    for more details).
+    deterministic order.
```
ZhitingHu: "pack in deterministic order" is a bit too vague. We can either make `pack_sequence_as` public in docs and refer to it, or copy docs of `pack_sequence_as` here to make it clear. (The former is probably preferred).

texar/torch/modules/pretrained/pretrained_base.py
```diff
@@ -181,8 +181,7 @@ def download_checkpoint(cls, pretrained_model_name: str,
         Args:
             pretrained_model_name (str): Name of the model checkpoint.
             cache_dir (str, optional): Path to the cache directory. If `None`,
-                uses the default directory given by
-                :meth:`~default_download_dir`.
+                uses the default directory (user's home directory).
```
ZhitingHu: Perhaps a better default directory is the current working directory instead of home?

huzecong: This was changed in issue #146 because we now use `pip install` without `-e`. `pip` would scan all files and having very large files under the project directory would make it unbearably slow (took a few minutes; I thought it froze at first). Other libraries (e.g. NLTK) also download stuff to a directory under home so I think it's okay. Plus, Texar-TF & Texar-PyTorch can share downloaded resources.

ZhitingHu: I see. Make sense

PR #165, opened 2019-08-19
Seperate RoBERTa from BERT

gpengzhi:
Since the tokenization procedures of `BERT` and `RoBERTa` are very different (`BERT` applies `wordpiece`, while `RoBERTa` applies `BPE`), we decide to separate `RoBERTa` from `BERT`.

As a result, we have `RoBERTaTokenizer` for `RoBERTa` models and `BERTTokenizer` for `BERT` model.

### DIFF
--- a/texar/torch/modules/classifiers/bert_classifier.py
+++ b/texar/torch/modules/classifiers/bert_classifier.py
@@ -24,6 +24,7 @@ from texar.torch.core.layers import get_initializer
 from texar.torch.hyperparams import HParams
 from texar.torch.modules.classifiers.classifier_base import ClassifierBase
 from texar.torch.modules.encoders.bert_encoder import BERTEncoder
+from texar.torch.modules.encoders.roberta_encoder import RoBERTaEncoder
 from texar.torch.modules.pretrained.pretrained_bert import PretrainedBERTMixin
 from texar.torch.utils.utils import dict_fetch
 
@@ -68,11 +69,21 @@ class BERTClassifier(ClassifierBase, PretrainedBERTMixin):
         super().__init__(hparams=hparams)
 
         # Create the underlying encoder
-        encoder_hparams = dict_fetch(hparams, BERTEncoder.default_hparams())
+        if self.__class__.__name__.startswith('BERT'):
+            encoder_hparams = dict_fetch(hparams, BERTEncoder.default_hparams())
 
-        self._encoder = BERTEncoder(pretrained_model_name=pretrained_model_name,
-                                    cache_dir=cache_dir,
-                                    hparams=encoder_hparams)
+            self._encoder = BERTEncoder(
+                pretrained_model_name=pretrained_model_name,
+                cache_dir=cache_dir,
+                hparams=encoder_hparams)
+        elif self.__class__.__name__.startswith('RoBERTa'):
+            encoder_hparams = dict_fetch(hparams,
+                                         RoBERTaEncoder.default_hparams())
+
+            self._encoder = RoBERTaEncoder(
+                pretrained_model_name=pretrained_model_name,
+                cache_dir=cache_dir,
+                hparams=encoder_hparams)
 
         # Create a dropout layer
         self._dropout_layer = nn.Dropout(self._hparams.dropout)
@@ -230,9 +241,14 @@ class BERTClassifier(ClassifierBase, PretrainedBERTMixin):
                   ``[batch_size, max_time, num_classes]`` and ``pred`` is of
                   shape ``[batch_size, max_time]``.
         """
-        enc_outputs, pooled_output = self._encoder(inputs,
-                                                   sequence_length,
-                                                   segment_ids)
+        if self.__class__.__name__.startswith('BERT'):
+            enc_outputs, pooled_output = self._encoder(inputs,
+                                                       sequence_length,
+                                                       segment_ids)
+        elif self.__class__.__name__.startswith('RoBERTa'):
+            enc_outputs, pooled_output = self._encoder(inputs,
+                                                       sequence_length)
+
         # Compute logits
         strategy = self._hparams.clas_strategy
         if strategy == 'time_wise':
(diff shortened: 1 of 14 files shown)


huzecong:
But I guess the classifier & encoder implementations are basically the same? Can we somehow unify those interfaces so we don't need to write the same code twice?

gpengzhi:
> But I guess the classifier & encoder implementations are basically the same? Can we somehow unify those interfaces so we don't need to write the same code twice?

If we want to reuse that part of the code, we probably should merge `PretrainedBERTMixin` and `PretrainedRoBERTaMixin` (?) Please let me know if there is a cleaner way to do this.

huzecong:
I'm thinking we can create `RoBERTaClassifier` by subclassing `PretrainedRoBERTaMixin` and `BERTClassifier`, this way the RoBERTa mixin class appears before BERT mixin in the MRO. I'm not sure if this would work, could you test?

gpengzhi:
MRO Order for `RoBERTaEncoder`: `RoBERTaEncoder` -> `PretrainedRoBERTaMixin` -> `BERTEncoder` -> `EncoderBase` -> `PretrainedBERTMixin` -> `PretrainedMixin` -> `ModuleBase` -> `ABC`

gpengzhi:
MRO Order for `RoBERTaClassifier`: `RoBERTaClassifier` -> `PretrainedRoBERTaMixin` -> `BERTClassifier` -> `ClassifierBase` -> `PretrainedBERTMixin` -> `PretrainedMixin` -> `ModuleBase` -> `ABC`

gpengzhi:
> I'm thinking we can create `RoBERTaClassifier` by subclassing (in this order) `PretrainedRoBERTaMixin` and `BERTClassifier`, this way the RoBERTa mixin class appears before BERT mixin in the MRO. I'm not sure if this would work, could you test?

Thanks for the help! I was too lazy to try Python Inheritance this afternoon...

texar/torch/modules/classifiers/bert_classifier.py
```diff
@@ -230,9 +241,14 @@ def forward(self,  # type: ignore
                   ``[batch_size, max_time, num_classes]`` and ``pred`` is of
                   shape ``[batch_size, max_time]``.
         """
-        enc_outputs, pooled_output = self._encoder(inputs,
-                                                   sequence_length,
-                                                   segment_ids)
+        if self.__class__.__name__.startswith('BERT'):
```
huzecong: Let's make this part a separate function instead. Conditioning on class name seems too hacky.

texar/torch/modules/classifiers/bert_classifier.py
```diff
@@ -68,11 +69,21 @@ def __init__(self,
...
+                cache_dir=cache_dir,
+                hparams=encoder_hparams)
+        elif self.__class__.__name__.startswith('RoBERTa'):
+            encoder_hparams = dict_fetch(hparams,
+                                         RoBERTaEncoder.default_hparams())
+
+            self._encoder = RoBERTaEncoder(
+                pretrained_model_name=pretrained_model_name,
+                cache_dir=cache_dir,
+                hparams=encoder_hparams)
```
huzecong: Same here. We can add an class-level attribute that stores the encoder class.

texar/torch/modules/encoders/bert_encoder.py
```diff
@@ -71,7 +71,8 @@ def __init__(self,
 
         # Segment embedding for each type of tokens
         self.segment_embedder = None
-        if self._hparams.type_vocab_size > 0:
+        if self._hparams.get('type_vocab_size') and \
```
huzecong: I believe this can be rewritten as:
```python
if self._hparams.get('type_vocab_size', 0) > 0:
```

texar/torch/modules/encoders/roberta_encoder.py
```diff
@@ -279,30 +239,8 @@ def forward(self,  # type: ignore
               character of the input (`CLS`), see RoBERTa's paper.
```
huzecong: Since RoBERTa encoder does not use `segment_ids`, let's mention it in the docstring.

texar/torch/modules/classifiers/bert_classifier.py
```diff
@@ -69,14 +69,14 @@ def __init__(self,
...
             encoder_hparams = dict_fetch(hparams, BERTEncoder.default_hparams())
 
             self._encoder = BERTEncoder(
                 pretrained_model_name=pretrained_model_name,
                 cache_dir=cache_dir,
                 hparams=encoder_hparams)
-        elif self.__class__.__name__.startswith('RoBERTa'):
+        elif self._MODEL_NAME == 'RoBERTa':
             encoder_hparams = dict_fetch(hparams,
                                          RoBERTaEncoder.default_hparams())
```
huzecong: Sorry but this is not what I meant. What I meant was something along the lines of:
```python
class BERTClassifier(...):
  _ENCODER_CLASS = BERTEncoder

  def __init__(...):
    ...
    encoder_hparams = dict_fetch(hparams, self._ENCODER_CLASS.default_hparams())
    ...

class RoBERTaClassifier(...):
  _ENCODER_CLASS = RoBERTaEncoder
```

texar/torch/modules/classifiers/bert_classifier.py
```diff
@@ -241,11 +241,11 @@ def forward(self,  # type: ignore
...
         """
-        if self.__class__.__name__.startswith('BERT'):
+        if self._MODEL_NAME == 'BERT':
             enc_outputs, pooled_output = self._encoder(inputs,
                                                        sequence_length,
                                                        segment_ids)
-        elif self.__class__.__name__.startswith('RoBERTa'):
+        elif self._MODEL_NAME == 'RoBERTa':
             enc_outputs, pooled_output = self._encoder(inputs,
                                                        sequence_length)
```
huzecong: And I don't think we need this check at all. Don't we ignore `segment_ids` in `RoBERTaEncoder`?

gpengzhi: `RoBERTaEncoder` only accepts two arguments (`inputs` and `sequence_length`). Do you mean we add `segment_ids` back in `RoBERTaEncoder` and set it always to be `None`?

huzecong: Since the classifier already does this, I think we can add it back.

huzecong: We could check that `segment_ids` is `None` and give a warning otherwise.

PR #168, opened 2019-08-23
Add basic version of tbX

AvinashBukkittu:
This is a basic version of tbX integration. It introduces the following changes

- The train metrics and validation metrics provided by the user are visualized as scalars on tensorboard.

- User can additionally supply frequency of logging for train metrics through `tbx_log_every` argument in the constructor

- The validation metrics are currently logged after every epoch.

### DIFF
--- a/texar/torch/run/executor.py
+++ b/texar/torch/run/executor.py
@@ -34,6 +34,8 @@ from torch import nn
 from torch.optim.lr_scheduler import _LRScheduler as LRScheduler
 from torch.optim.optimizer import Optimizer
 
+from tensorboardX import SummaryWriter
+
 from texar.torch.data.data.data_base import DataBase
 from texar.torch.data.data.data_iterators import BatchingStrategy, DataIterator
 from texar.torch.data.data.dataset_utils import Batch
@@ -714,6 +716,9 @@ class Executor:
                  test_data: OptionalDict[DataBase] = None,
                  batching_strategy: Optional[BatchingStrategy] = None,
                  device: Optional[torch.device] = None,
+                 # tbX logging
+                 tbx_logging_dir: Optional[str] = None,
+                 tbx_log_every: Optional[Condition] = None,
                  # Checkpoint
                  checkpoint_dir: Optional[str] = None,
                  max_to_keep: Optional[int] = None,
@@ -760,7 +765,6 @@ class Executor:
         except ImportError:
             self._tty_ncols = None
 
-        # TODO: Add support for Tensorboard.
         # Meta variables
         self._should_terminate = False  # .terminate()
         self._should_remove_current_action = False  # .remove_action()
@@ -900,6 +904,12 @@ class Executor:
                for stage in show_live_progress):
             raise ValueError("Invalid value for `show_live_progress`")
         self._register_logging_actions(show_live_progress)
+
+        # tbx logging
+        if tbx_logging_dir is not None:
+            self.summary_writer = SummaryWriter(logdir=tbx_logging_dir)
+            self._tbx_logging_conditions = utils.to_list(tbx_log_every)
+            self._register_tbx_logging_actions()
 
         # Register other events and actions.
         for cond in self._valid_conditions:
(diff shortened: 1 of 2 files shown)


texar/torch/run/executor.py
```diff
@@ -901,6 +905,12 @@ def __init__(self, model: nn.Module,
             raise ValueError("Invalid value for `show_live_progress`")
         self._register_logging_actions(show_live_progress)
 
+        # tbx logging
+        if tbx_logging_dir is not None:
+            self.summary_writer = SummaryWriter(logdir=tbx_logging_dir)
+            self._tbx_logging_conditions = utils.to_list(tbx_log_every)
```
huzecong: Can we resort to `log_every` if `tbx_log_every` is `None`?

texar/torch/run/executor.py
```diff
@@ -1483,6 +1493,37 @@ def erase_status_and_log_fn(executor: 'Executor'):
...
+
+        def tbx_eval_log_fn(executor: 'Executor'):
+            valid_metrics = executor.valid_metrics
+
+            for key, value in valid_metrics.items():
+                self.summary_writer.add_scalar(
+                    f"eval/{key}", value.value(), executor.status["epoch"])
+
+        _register(self._tbx_logging_conditions, tbx_train_log_fn)
+        _register([Event.Epoch], tbx_eval_log_fn)
```
huzecong: Validation is performed on `validation_every` conditions, it's not necessarily per epoch. It's also possible to perform validation multiple times in an epoch (e.g., the Transformer example). I think we can also use iteration as the X-axis value.

huzecong: And, instead of `eval`, maybe we can just call it `valid`. To me `eval` feels like the union of valid and test.

texar/torch/run/executor.py
```diff
@@ -1483,6 +1493,37 @@ def erase_status_and_log_fn(executor: 'Executor'):
...
+            for key, value in train_metrics.items():
+                self.summary_writer.add_scalar(
+                    f"train/{key}", value.value(), executor.status["iteration"])
+
+        def tbx_eval_log_fn(executor: 'Executor'):
+            valid_metrics = executor.valid_metrics
+
+            for key, value in valid_metrics.items():
+                self.summary_writer.add_scalar(
+                    f"eval/{key}", value.value(), executor.status["epoch"])
```
huzecong: A potential problem here is `metric.value()` will be called twice after each validation (once for logging, once for tbX). For `StreamingMetric` this is not a problem, but for `SimpleMetric` this may be time consuming. Maybe we can modify the interface for `SimpleMetric` a bit, e.g., we can cache the previous result calculated in `value()`, until new values are added or `reset()` is called. This caching routine could be performed in the `SimpleMetric` class `value()` method, and subclasses of `SimpleMetric` would then need to implement another method (e.g. `_value()`) instead of `value()`.

Basically:

```python
class SimpleMetric(Metric[Input, Value]):
  _cached_value: Optional[Value]

  def reset(self):
    self._cached_value = None

  def add(...):
    self._cached_value = None

  def value(self):
    if self._cached_value is not None:
      return self._cached_value
    self._cached_value = self._value()
    return self._cached_value

  def _value(self) -> Value:
    r""" What previously goes in `value()` now goes here. """
    raise NotImplementedError
```

texar/torch/run/executor.py
```diff
@@ -714,6 +716,9 @@ def __init__(self, model: nn.Module,
                  test_data: OptionalDict[DataBase] = None,
                  batching_strategy: Optional[BatchingStrategy] = None,
                  device: Optional[torch.device] = None,
+                 # tbX logging
+                 tbx_logging_dir: Optional[str] = None,
+                 tbx_log_every: Optional[Condition] = None,
```
huzecong: The original tensorboard arguments at the bottom can safely go away now, I think.

huzecong: Also, do we support setting experiment names now?

huzecong:
The unit tests are not run in CI because tbX is not installed. Please also add tbX as a requirement under `extras_require['extras']` in `setup.py`.

texar/torch/run/executor.py
huzecong: Not really. The iteration counts from the beginning of training and does not reset after each epoch. So if `executor.status["iteration"] == 10000` after the first epoch, it will become `30000` after the third epoch.

You can also add entries to the `status` dictionary as you wish, just make sure you document it. But I feel it's better to use iteration rather than validation counts, because it's possible that validation is performed at non-uniform intervals.

texar/torch/run/executor.py
AvinashBukkittu: Resolved in bbaf278

texar/torch/run/executor.py
AvinashBukkittu: Fixed in bbaf278

texar/torch/run/executor.py
AvinashBukkittu: Currently, we don't. TBX logs are written directly to `tbx_logging_dir`. Removed tensorboard args in bbaf278

texar/torch/run/executor.py
```diff
@@ -332,6 +332,17 @@ def forward(self, batch):
...
+    .. _executor-tbx-logging-args:
+
+    **Arguments for tensorboardX logging:**
+
+    `tbx_logging_dir`: str
+        Path to the directory for storing tensorboard logs.
+
+    `tbx_log_every`: |Condition|
+        Conditions that, when triggered, saves the tensorboard logs for train
+        metrics.
```
huzecong: Please mention that `log_every` will be used if `None` is provided. Also, the `.. _` marks an in-page anchor, please add an entry to the list of grouped parameters (here: https://github.com/asyml/texar-pytorch/blob/bbaf278e19ce153bb90546b47e9c9d0dbc475fca/texar/torch/run/executor.py#L256-L263).

texar/torch/run/metric/base_metric.py
```diff
@@ -150,17 +150,29 @@ class SimpleMetric(Metric[Input, Value], ABC):
...
+        self._cached_value = None
+
+    def value(self):
+        if self._cached_value is not None:
+            return self._cached_value
+        self._cached_value = self._value()
+        return self._cached_value
+
+    def _value(self) -> Value:
+        raise NotImplementedError
```
huzecong: Please also document this behavior so users understand.

AvinashBukkittu:
> Can we add an example (or update an existing one) to showcase the use of the tbX feature ?

I will show tbX feature in BERT classifier example.

PR #170, opened 2019-08-23
Update GPT-2 Model List

### LINKED ISSUE #169: Support GPT-2 774M checkpoint file & Rename the current GPT2 checkpoint files
https://github.com/openai/gpt-2

In their official website, `Note that our original parameter counts were wrong due to an error (in our previous blog posts and paper). Thus you may have seen small referred to as 117M and medium referred to as 345M.`

We should rename ours as well. `We have currently released small (124M parameter), medium (355M parameter), and large (774M parameter) versions of GPT-2.`

gpengzhi:
Resolve #169

### DIFF
--- a/texar/torch/modules/pretrained/pretrained_gpt2.py
+++ b/texar/torch/modules/pretrained/pretrained_gpt2.py
@@ -70,11 +70,18 @@ class PretrainedGPT2Mixin(PretrainedMixin, ABC):
             if key == '117M':
                 warnings.warn("Pre-trained model name '117M' is deprecated, "
                               "use 'gpt2-small' instead.", UserWarning)
+                return True
             elif key == '345M':
                 warnings.warn("Pre-trained model name '345M' is deprecated, "
                               "use 'gpt2-medium' instead.", UserWarning)
+                return True
             else:
                 return super().__contains__(key)
+    _DEPRECATED_MODEL2URL = {
+        '117M': [_GPT2_PATH + f"124M/{file}" for file in _CHECKPOINT_FILES],
+        '345M': [_GPT2_PATH + f"355M/{file}" for file in _CHECKPOINT_FILES],
+    }
+    _MODEL2URL.update(_DEPRECATED_MODEL2URL)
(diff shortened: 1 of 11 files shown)


texar/torch/modules/pretrained/pretrained_gpt2.py
```diff
@@ -46,18 +46,36 @@ class PretrainedGPT2Mixin(PretrainedMixin, ABC):
...
+                        for file in _CHECKPOINT_FILES],
+        'gpt2-large': [_GPT2_PATH + f"774M/{file}"
+                       for file in _CHECKPOINT_FILES],
     }
 
+    # Raise warning for the deprecated pre-trained model names
+    class MyDict(dict):
+        def __contains__(self, key):
+            if key == '117M':
+                raise KeyError("Pre-trained model name '117M' is deprecated, "
```
huzecong: Raising a `KeyError` is not really a "deprecation warning". Let's user the `warnings` module for that.

gpengzhi:
> `raise`ing anything results in exceptions being thrown. You need `warnings.warn` and specify the type `UserWarning` there.

I see. Maybe we can use `DeprecationWarning` instead?

huzecong:
`DeprecationWarning` is ignored by default, and they're intended towards developers. Basically, any warning that we intend the end-user to see should be `UserWarning`. Please refer to https://docs.python.org/3/library/warnings.html#warning-categories

gpengzhi:
> `DeprecationWarning` is ignored by default, and they're intended towards developers. Basically, any warning that we intend the end-user to see should be `UserWarning`. Please refer to https://docs.python.org/3/library/warnings.html#warning-categories

Thanks for the explanation!

huzecong:
Now the user will get a warning when using `115M` or `337M`, but they can't really use it right? `__contains__` returns `None` for these two keys, and there's actually no values corresponding to them.

gpengzhi:
I must misunderstand the definition of warning. I will modify it accordingly.

gpengzhi:
My bad. I thought there is no big difference between raising exceptions and raising warnings.

PR #181, opened 2019-08-31
Move regex and sentencepiece from extras_require to install_requires

gpengzhi:
Our `Tokenizer` modules have a dependency on these two packages. `regex` and `sentencepiece` should be moved to `install_requires`. Otherwise, `ModuleNotFoundError` will be triggered (if `regex` and `sentencepiece` are not installed) when they try to `import texar.torch`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,3 +2,5 @@ torch >= 1.0.0
 funcsigs >= 1.0.2
 numpy >= 1.15.4
 mypy_extensions >= 0.4.1
+regex
+sentencepiece
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,5 +2,5 @@ torch >= 1.0.0
 funcsigs >= 1.0.2
 numpy >= 1.15.4
 mypy_extensions >= 0.4.1
-regex
-sentencepiece
+regex >= 2018.01.10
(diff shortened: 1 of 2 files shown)


ZhitingHu:
good catch. we may wanna update `requirements.txt` accordingly

gpengzhi:
I noticed that we put `torch` in `extra_require` as well. So it may be ok to leave `regex` and `sentencepiece` as they are (?) @huzecong

gpengzhi:
> I noticed that we put `torch` in `extra_require` as well. So it may be ok to leave `regex` and `sentencepiece` as they are (?) @huzecong

If that is the case, we might need to mention that users need to install `regex` and `sentencepiece` to use our package in `README.md`.

huzecong:
I guess it's fair to have these in `install_requires`, because tokenizers is a feature that most people would use (as pre-trained models depend on them). This is contrary to `Pillow` which is only used if you load images with `RecordData`.

The reason we don't have `torch` in the list is because there are multiple version of PyTorch that goes with different CUDA versions (or CPU only). Instead of installing for the user and fail, it might be better to leave it to the user themselves.

huzecong:
Can we also add a version limit for the new requirements? And in `setup.py` as well?

PR #187, opened 2019-09-04
Fix potential position embedding issue in decoders

### LINKED ISSUE #158: Potential position embedding issue in decoders.
Currently there is a potential issue in some decoders(e.g. `AttentionRNNDecoder`) which support position embedder. 

In eval mode, given `max_decoding_length`, the `dynamic_decode` will stop at `time==max_decoding_length - 1`, while decoder helper may pass `time = time + 1` to its position embedder. Since `position_size` of the position embedder  is usually equal to `max_decoding_length`, inputing `time==max_decoding_length` will cause `index out of range` error on position embedder.

Suggested by @huzecong , _we could fix this for all decoders by adding a check in the embedding function._

gpengzhi:
Resolve #158

### DIFF
--- a/texar/torch/modules/decoders/decoder_base.py
+++ b/texar/torch/modules/decoders/decoder_base.py
@@ -27,6 +27,7 @@ from texar.torch.module_base import ModuleBase
 from texar.torch.modules.decoders import decoder_helpers as helpers
 from texar.torch.modules.decoders.decoder_helpers import Helper
 from texar.torch.utils import utils
+from texar.torch.utils.dtypes import torch_bool
 
 __all__ = [
     '_make_output_layer',
@@ -405,8 +406,17 @@ class DecoderBase(ModuleBase, Generic[State, Output], ABC):
 
         while (not torch.all(finished).item() and
                (max_decoding_length is None or time < max_decoding_length)):
-            (next_outputs, decoder_state, next_inputs,
-             decoder_finished) = self.step(helper, time, step_inputs, state)
+
+            next_outputs, decoder_state = \
+                self.compute_output_state(helper, time, step_inputs, state)
+
+            if max_decoding_length is not None and \
+                    time + 1 == max_decoding_length:
+                next_inputs = torch.empty(0)
+                decoder_finished = torch.tensor(1, dtype=torch_bool)
+            else:
+                next_inputs, decoder_finished = \
+                    self.compute_next_input(helper, time, next_outputs)
 
             if getattr(self, 'tracks_own_finished', False):
                 next_finished = decoder_finished
@@ -480,10 +490,11 @@ class DecoderBase(ModuleBase, Generic[State, Output], ABC):
         raise NotImplementedError
 
     @abstractmethod
-    def step(self, helper: Helper, time: int,
-             inputs: torch.Tensor, state: Optional[State]) \
-            -> Tuple[Output, State, torch.Tensor, torch.ByteTensor]:
-        r"""Called per step of decoding (but only once for dynamic decoding).
+    def compute_output_state(self, helper: Helper, time: int,
+                             inputs: torch.Tensor, state: Optional[State]) \
+            -> Tuple[Output, State]:
+        r"""Compute the output and the state at the current time step.
+        Called per step of decoding (but only once for dynamic decoding).
 
         Args:
             helper: The :class:`~texar.torch.modules.Helper` instance to use.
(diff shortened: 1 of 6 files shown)


ZhitingHu:
Thanks for the quick update. This is a critical fix that can potentially impact many modules examples. Can you add more unit tests (if applicable) to make sure it works as intended, and run relevant examples (e.g., `seq2seq_attn`, `transformer`) to see if the results can be reproduced?

texar/torch/modules/decoders/decoder_base.py
```diff
@@ -405,8 +406,22 @@ def dynamic_decode(self, helper: Helper, inputs: Optional[torch.Tensor],
...
+
+            next_outputs, decoder_state = \
+                self.compute_output_state(helper, time, step_inputs, state)
+
+            if max_decoding_length is not None and \
+                    time + 1 == max_decoding_length:
+                # Maximum decoding length reached, mark all batches as finished.
+                # This requires special handling because performing lookup on
+                # position embeddings with `time + 1` may result in IndexError.
+                decoder_finished = torch.tensor(1, dtype=torch_bool)
```
huzecong: This should probably be put on the same device as other tensors. Have you tested the code on GPU?

texar/torch/modules/decoders/xlnet_decoder.py
```diff
@@ -281,10 +279,15 @@ def step(self,
         logits = F.linear(net_output, self.word_embed.weight, self.lm_bias)
         logits = logits[:, -1]
         sample_ids = helper.sample(time=time, outputs=logits)
-        (finished, next_inputs) = helper.next_inputs(
-            self.embed_tokens, time, logits, sample_ids)
         outputs = XLNetDecoderOutput(logits=logits, sample_id=sample_ids)
-        return outputs, memory, next_inputs, finished
+        return outputs, memory
+
+    def compute_next_input(self, helper: Helper, time: int,
```
huzecong: Is `compute_next_input` identical for all the derived classes? If it is, let's make it the default impl. in the base class.

texar/torch/modules/decoders/decoder_base.py
gpengzhi: good point. I think it should have the same `device` as that of `finished`.

texar/torch/modules/decoders/decoder_base.py
```diff
@@ -408,20 +408,23 @@ def dynamic_decode(self, helper: Helper, inputs: Optional[torch.Tensor],
...
+                decoder_finished = torch.tensor(1, dtype=torch_bool,
+                                                device=finished.device)
                 # Since `next_inputs` will not be used, simply create a null
                 # tensor.
                 next_inputs = torch.empty(0)
             else:
-                next_inputs, decoder_finished = \
-                    self.compute_next_input(helper, time, next_outputs)
+                logits = next_outputs.logits  # type: ignore
+                sample_id = next_outputs.sample_id  # type: ignore
```
huzecong: Why do we need `type: ignore` here?

gpengzhi: `next_outputs` is an `Output` object, and it has no attributes `logits` and `sample_id`.

gpengzhi:
> Thanks for the quick update. This is a critical fix that can potentially impact many modules examples. Can you add more unit tests (if applicable) to make sure it works as intended, and run relevant examples (e.g., `seq2seq_attn`, `transformer`) to see if the results can be reproduced?

Tested `seq2seq_attn` and `transformer` on `GPU`. Results can be reproduced.

gpengzhi:
> Thanks for the quick update. This is a critical fix that can potentially impact many modules examples. Can you add more unit tests (if applicable) to make sure it works as intended, and run relevant examples (e.g., `seq2seq_attn`, `transformer`) to see if the results can be reproduced?

Unit tests in `transformer_decoders_test.py` have already tested such behavior.

texar/torch/modules/decoders/decoder_base.py
```diff
@@ -405,8 +406,25 @@ def dynamic_decode(self, helper: Helper, inputs: Optional[torch.Tensor],
...
+                # position embeddings with `time + 1` may result in IndexError.
+                decoder_finished = torch.tensor(1, dtype=torch_bool,
+                                                device=finished.device)
+                # Since `next_inputs` will not be used, simply create a null
+                # tensor.
+                next_inputs = torch.empty(0)
+            else:
+                logits = next_outputs.logits  # type: ignore
+                sample_id = next_outputs.sample_id  # type: ignore
+                decoder_finished, next_inputs = helper.next_inputs(
```
ZhitingHu: It'd be better to add a `next_inputs` method to the decoder and call here (instead of calling helper's `next_inputs` directly). 

One reason is that in TF, `dynamic_decode` (https://github.com/tensorflow/tensorflow/blob/r1.14/tensorflow/contrib/seq2seq/python/ops/decoder.py#L268-L472) does not have access to helper, but only decoder. Though we will also update TF `dynamic_decode` to add the `if.. else..` block like here, it's good to not change the factorization in TF (helper -- decoder -- `dynamic_decode`, where `dynamic_decode` does not operate helper directly). To make TF/Pytorch have the similar structures, let's add `next_inputs` to the decoders.

huzecong:
Wait... I understand the need to add back `next_inputs`, but why can't we provide a default implementation in the base class? It's still the same in every derived class, just as before.

ZhitingHu:
I agree. Let's remove the duplicate code

gpengzhi:
I will remove the duplicate code.

Issue #188, opened 2019-09-04
MultiAlignedData cannot correctly return int64 data


gpengzhi:
1. Specify [data_type](https://github.com/asyml/texar-pytorch/blob/master/examples/sentence_classifier/config_kim.py#L45) to be `int64`.
2. Delete [.long()](https://github.com/asyml/texar-pytorch/blob/master/examples/sentence_classifier/clas_main.py#L61)
3. Run `sentence_classifier` example.

```
ValueError: 'int64' is not a valid _DataType
```

A related issue with closed #172

gpengzhi:
Resolved in #189

Issue #173, opened 2019-08-26
Support torch.int64 in ScalarData


gpengzhi:
In our implementation of `ScalarData`, we only support `torch.int32` data. However, if the scalar data is `label` data, and we want to use it to compute loss (for example, `F.cross_entropy(input, target)`), we have to manually convert it to `torch.int64`, otherwise an error will be triggered. 
```
RuntimeError: Expected object of scalar type Long but got scalar type Int for argument #2 'target'
```

PR #189, opened 2019-09-05
Add additional datatype support for ScalarData

AvinashBukkittu:
This PR

- Adds additional datatype support for `ScalarData` in `MultiAlignedData` (We now have `get_supported_scalar_types()` method in `dtypes.py` to get list of supported data types.)

- Adds test cases for additional scalar data types supported.

- Adds negative test cases for unsupported scalar data types.

### DIFF
--- a/texar/torch/data/data/multi_aligned_data.py
+++ b/texar/torch/data/data/multi_aligned_data.py
@@ -46,7 +46,9 @@ class _DataType(Enum):
     """
     TEXT = "text"
     INT = "int"
+    INT64 = "int64"
     FLOAT = "float"
+    FLOAT64 = "float64"
     RECORD = "record"
 
 
@@ -55,7 +57,8 @@ def _is_text_data(data_type):
 
 
 def _is_scalar_data(data_type):
-    return data_type in [_DataType.INT, _DataType.FLOAT]
+    return data_type in [_DataType.INT, _DataType.INT64, _DataType.FLOAT,
+                         _DataType.FLOAT64]
 
 
 def _is_record_data(data_type):
--- a/texar/torch/data/data/multi_aligned_data.py
+++ b/texar/torch/data/data/multi_aligned_data.py
@@ -33,7 +33,7 @@ from texar.torch.data.embedding import Embedding
 from texar.torch.data.vocabulary import SpecialTokens, Vocab
 from texar.torch.hyperparams import HParams
 from texar.torch.utils import utils, dict_fetch
-from texar.torch.utils.dtypes import is_str
+from texar.torch.utils.dtypes import is_str, get_supported_scalar_types
 
 __all__ = [
     "_default_dataset_hparams",
@@ -41,28 +41,16 @@ __all__ = [
 ]
 
 
-class _DataType(Enum):
-    r"""Enumeration of data types.
-    """
-    TEXT = "text"
-    INT = "int"
-    INT64 = "int64"
-    FLOAT = "float"
-    FLOAT64 = "float64"
-    RECORD = "record"
-
-
 def _is_text_data(data_type):
-    return data_type is _DataType.TEXT
+    return data_type == "text"
 
 
 def _is_scalar_data(data_type):
-    return data_type in [_DataType.INT, _DataType.INT64, _DataType.FLOAT,
-                         _DataType.FLOAT64]
+    return data_type in get_supported_scalar_types()
 
 
 def _is_record_data(data_type):
-    return data_type is _DataType.RECORD
+    return data_type == "record"
 
 
 def _default_dataset_hparams(data_type=None):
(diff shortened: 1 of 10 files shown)


huzecong:
Why don't we just follow #173?

gpengzhi:
@AvinashBukkittu Before this one is merged, can you modify `sentence_classifier` example by following the steps in #188 to make sure this is actually working?

texar/torch/data/data/multi_aligned_data_test.py
```diff
@@ -203,6 +203,29 @@ def test_length_filter(self):
              "length_filter_mode": "truncate"})
         self._run_and_test(hparams, discard_index=0)
 
+    def test_scalar_types(self):
```
huzecong: Might as well test that we correctly reject non-supported types.

texar/torch/utils/dtypes.py
```diff
@@ -96,6 +87,16 @@ def is_str(x):
     return isinstance(x, str)
 
 
+def get_supported_scalar_types():
```
huzecong: Oh, I thought we were only going to support a part of these types. For instance, `np.str` doesn't seem like it should be in `ScalarData`.

texar/torch/data/data/multi_aligned_data.py
```diff
@@ -68,13 +58,12 @@ def _default_dataset_hparams(data_type=None):
     See :meth:`texar.torch.data.MultiAlignedData.default_hparams` for details.
     """
     if data_type is None:
-        data_type = _DataType.TEXT
-    else:
-        data_type = _DataType(data_type)
+        data_type = "text"
```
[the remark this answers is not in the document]
AvinashBukkittu: Here https://github.com/asyml/texar-pytorch/blob/master/texar/torch/data/data/multi_aligned_data.py#L184 we pass `None`. I thought I will the keep it as is.

texar/torch/data/data/multi_aligned_data_test.py
AvinashBukkittu: Added in 0758eb5. Added tests in `scalar_data_test.py` as well

texar/torch/utils/dtypes.py
AvinashBukkittu: Resolved in 0758eb5

texar/torch/data/data/scalar_data.py
```diff
@@ -162,7 +165,15 @@ def default_hparams():
 
     def process(self, raw_example: List[str]) -> Union[int, float]:
         assert len(raw_example) == 1
-        example: Union[int, float] = self._data_type(raw_example[0])
+
+        example_: Union[int, str]
+        if self._data_type == np.bool_:
+            example_ = int(raw_example[0])
```
huzecong: Why is it necessary to convert boolean values into `int`?

AvinashBukkittu: raw_example that is read from file is of type `str`, like "0", and `np.bool("0")` returns `True`. Hence, I convert it to `int` and `np.bool(0)` returns `False`

AvinashBukkittu:
> @AvinashBukkittu Before this one is merged, can you modify `sentence_classifier` example by following the steps in #188 to make sure this is actually working?

Yes, I tried the example and it seems to run.

texar/torch/data/data/multi_aligned_data.py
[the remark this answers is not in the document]
huzecong: I see. Yes, let's keep it as is.

texar/torch/data/data/scalar_data.py
huzecong: But it's possible to have strings like `"true"` which would fail here. I recommend using [`distutils.util.strtobool`](https://docs.python.org/3.7/distutils/apiref.html?highlight=distutils.util#distutils.util.strtobool).

texar/torch/data/data/scalar_data_test.py
```diff
@@ -68,18 +86,23 @@ def _run_and_test(self, hparams, test_transform=False):
...
+            data_type = hparams["dataset"]["data_type"]
+            if data_type == "int":
                 self.assertTrue(isinstance(value, torch.Tensor))
                 self.assertEqual(value.dtype, torch.int32)
-            else:
+            elif data_type == "float":
                 self.assertTrue(isinstance(value, torch.Tensor))
                 self.assertEqual(value.dtype, torch.float32)
+            elif data_type == "bool":
+                self.assertTrue(isinstance(value, torch.Tensor))
```
huzecong: A bit unrelated, but we can use `assertIsInstance` here, and since this check exists in all branches, we can move it outside.

AvinashBukkittu: 52fca5f

texar/torch/data/data/scalar_data.py
AvinashBukkittu: 52fca5f

PR #193, opened 2019-09-06
Resolve comments in #187

gpengzhi:


texar/torch/modules/decoders/decoder_base.py
```diff
@@ -535,7 +534,11 @@ def next_inputs(self, helper: Helper, time: int, outputs: Output) -> \
             - ``finished`` is a :torch:`ByteTensor` tensor telling whether the
               sequence is complete, for each sequence in the batch.
         """
-        raise NotImplementedError
+        logits = outputs.logits  # type: ignore
+        sample_id = outputs.sample_id  # type: ignore
```
huzecong: Looks good overall. What was the reason for these `type: ignore` comments?

gpengzhi: In `decoder_base.py`, outputs is an `Output` (`Output = TypeVar('Output')`) which does not have attributes `logits` and `sample_id`.

huzecong: I see. I feel kind of weird requiring `outputs` of derived decoders to be a structure that must contain `logits` and `sample_id` (although it is the case for our current ones). I also realized that `logits` were used in none of the helpers (the only usage is in TF `ScheduledOutputsTrainingHelper` where the output is directly fed into the model as the next input by some probability); while `sample_id` is just the value computed using the helper. Do you think maybe we can further refactor the code to make it even simpler?

gpengzhi: I agree. This might not be a good and safe coding style in the base class. I can refactor it a little bit to get rid of such an issue.

gpengzhi: I feel that it is reasonable for `step` function to return `wrapper_output` and `wrapper_state` (Actually I think we should only use wrapper level output and state instead of `logits` and `sample_ids` in `DecoderBase`). So we probably should call `helper.sample()` in `step` function. I do not want to change the interface of `Helper.next_input()` since `logits` might be used in the new type of `Helper` in the future. I am thinking of:
1. Add attribute checking in `next_input()` in `DeocderBase`.
or
2. Just define `next_input()` in every decoder class even though this introduces the code redundancy.

Please let me know your thoughts. @huzecong

huzecong: I agree with you on keeping the signature of `step` and `Helper` methods. Of the two proposals, I actually prefer the second one, because attribute checking sounds weird (and what happens when there's no such attribute? If an exception is thrown, would it be the same as it is now?).
I guess this means we'll be closing the PR for now. Sorry for wasting your efforts.

gpengzhi: No worries. Glad to improve the code quality of our repo.

Issue #199, opened 2019-09-09
Update the docs of Connectors module


gpengzhi:
Requested in #198

PR #162, opened 2019-08-19
Update connector

TomNong:
* Update `_mlp_transform` method.
* Polish connector.

PR #200, opened 2019-09-09
Fix connector docs

AvinashBukkittu:
This PR

- Uses `namedtuple('LSTMStateTuple', ['h', 'c'])(...)` instead of `namedtuple('LSTMStateTuple', ['c', 'h'])(...)` in the docs to avoid confusion (#199 )

- Addresses the comments of https://github.com/asyml/texar-pytorch/pull/162 and updates the docs.

### DIFF
--- a/texar/torch/modules/connectors/connectors.py
+++ b/texar/torch/modules/connectors/connectors.py
@@ -51,9 +51,9 @@ def _assert_same_size(outputs: TensorStruct,
     r"""Check if outputs match output_size
 
     Args:
-        outputs: A ``Tensor`` or a (nested) ``tuple`` of tensors
-        output_size: Can be an Integer, a ``torch.Size``, or a (nested)
-            ``tuple`` of Integers or ``torch.Size``.
+        outputs: A tensor or a (nested) tuple of tensors
+        output_size: Can be an ``int``, a ``torch.Size``, or a (nested)
+            tuple of ``int`` or ``torch.Size``.
     """
     flat_output_size = nest.flatten(output_size)
     flat_output = nest.flatten(outputs)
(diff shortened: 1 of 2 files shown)


texar/torch/modules/connectors/connectors.py
```diff
@@ -69,80 +69,74 @@ def _assert_same_size(outputs: TensorStruct,
...
     r"""Transforms inputs through a fully-connected layer that creates
     the output with specified size.
 
     Args:
-        inputs: A ``Tensor`` of shape ``[batch_size, ..., finale_state]``
+        inputs: A tensor of shape ``[batch_size, d1, ..., dn]``
             (i.e., batch-major), or a (nested) tuple of such elements.
-            A Tensor or a (nested) tuple of Tensors with shape
+            A tensor or a (nested) tuple of tensors with shape
             ``[max_time, batch_size, ...]`` (i.e., time-major) can
```
ZhitingHu: Is this PR include all changes in https://github.com/asyml/texar-pytorch/pull/162? It looks not to me. E.g., the correct doc of `inputs` here should be: https://github.com/asyml/texar-pytorch/pull/162/files#diff-4a3f72feda3ae73445008a08ee31878aR99

AvinashBukkittu: Added this sentence - `The dimensions ``d1, ..., dn`` will be flattened and transformed by a dense layer.` I retain

```
inputs: A tensor of shape ``[batch_size, d1, ..., dn]``
            (i.e., batch-major), or a (nested) tuple of such elements.
            A tensor or a (nested) tuple of tensors with shape
            ``[max_time, batch_size, ...]`` (i.e., time-major) can
            be transposed to batch-major using
            :func:`~texar.torch.utils.transpose_batch_time` prior to this
            function.
```

in this PR (which was deleted in #162) as I think this is more informative. #162 requested for 3 changes - one was this doc change, other two requested to add `_get_sizes` method in this file.

ZhitingHu: The original doc `A tensor or a (nested) tuple of tensors with shape [max_time, batch_size, ...] (i.e., time-major) ...` is incorrect. Let's copy the new docs from https://github.com/asyml/texar-pytorch/pull/162 directly. 

It corresponds to TF: https://github.com/asyml/texar/blob/master/texar/tf/modules/connectors/connectors.py#L78

AvinashBukkittu: Fixed in 570c7ab

texar/torch/modules/connectors/connectors.py
```diff
@@ -339,11 +322,10 @@ def forward(self,  # type: ignore
         Args:
             inputs: The input (structure of) tensor to pass forward.
 
-        :returns:
+        Returns:
```
huzecong: Using `:returns:` here is necessary because we have colons (`:`) in the paragraph. The `sphinx-napoleon` package would incorrectly interpret this as annotating a return type. Same for a bunch of docstrings below.

AvinashBukkittu: Fixed in 14ee0df

Issue #185, opened 2019-09-03
Support DistilBERT


gpengzhi:
https://medium.com/huggingface/distilbert-8cf3380435b5

PR #201, opened 2019-09-09
Add DistilBERT Modules

gpengzhi:
Knowledge distillation is a compression technique in which a small model (`DistilBERT`) is trained to reproduce the behavior of a larger model (`BERT`). In the teacher-student training, they train a student network (`DistilBERT`) to mimic the full output distribution of the teacher network (`BERT`).

The distillation training procedure is summarized as follows.

```python
import torch
import torch.nn as nn
import torch.nn.functional as F
from torch.optim import Optimizer

KD_loss = nn.KLDivLoss(reduction='batchmean')

def kd_step(teacher: nn.Module,
            student: nn.Module,
            temperature: float,
            inputs: torch.tensor,
            optimizer: Optimizer):
    teacher.eval()
    student.train()
    
    with torch.no_grad():
        logits_t = teacher(inputs=inputs)
    logits_s = student(inputs=inputs)
    
    loss = KD_loss(input=F.log_softmax(logits_s/temperature, dim=-1),
                   target=F.softmax(logits_t/temperature, dim=-1))
    
    loss.backward()
    optimizer.step()
    optimizer.zero_grad()
```

The architecture of `DistilBERT` is very similar to that of `BERT`, except:

`DistilBERT` reduces the number of layers by a factor of two.

`DistilBERT` removes the next sentence prediction objective so that there is no `segment_embedding` and `pooler` in the pre-training procedure.

`DistilBERT` uses sinusoidal position embedding during training, and such an embedding matrix is initialized from `bert-base-uncased`.


gpengzhi:
Resolve #185

texar/torch/data/tokenizers/pretrained_distilbert_tokenizer.py
```diff
@@ -0,0 +1,165 @@
...
+
+        `"non_split_tokens"`: list
+            List of tokens which will never be split during tokenization.
+            Only has an effect when `do_basic_tokenize=True`
+
+        `"name"`: str
+            Name of the tokenizer.
+
+        """
+        return {
```
ZhitingHu: Is `default_hparams` identical to `BertTokenizer`? If so, reuse by 
```
default_ = BertTokenzier.default_hparams()
default_['pretrained_model_name'] = 'distilbert-base-uncased'
return default_
```

If not, the statement in https://github.com/asyml/texar-pytorch/pull/201/files#diff-5910de50635db90a6aec9eb7ad7bbbc2R29 should be updated and point out which part is different

texar/torch/modules/encoders/distilbert_encoder.py
```diff
@@ -0,0 +1,254 @@
...
+    PretrainedDistilBERTMixin
+
+__all__ = [
+    "DistilBERTEncoder",
+]
+
+
+class DistilBERTEncoder(PretrainedDistilBERTMixin, BERTEncoder):  # type: ignore
+    r"""DistilBERT Transformer for encoding sequences.
+
```
ZhitingHu: Add a sentence: "Please see :class:`~texar.torch.modules.PretrainedDistilBERTaMixin` for a brief description of DistilBERT".

All other pretrained modules should add such a sentence

gpengzhi:
> Have we tested the accuracy of the DistilBERT modules?
> 
> If not, we can probably just push it back (leave the PR unmerged). TF pretrained modules are of higher priority

It is reported that DistilBERT has 5% of degradation from BERT. We have not tested it yet. I will leave this PR as it is, for now, this week.

PR #202, opened 2019-09-10
Minor fix

gpengzhi:


huzecong:
Hmm. So this is just a name change? I don't really feel it's necessary -- in fact, I feel having a `pretrained_` prefix for everything that's already under a `pretrained` module is a bit redundant. @ZhitingHu Do you think we can remove the `pretrained_` prefix for everything except `pretrained_base`, e.g.:

- `pretrained_bert` to `bert`
- `pretrained_xlnet_test` to `xlnet_test`

gpengzhi:
> Hmm. So this is just a name change? I don't really feel it's necessary -- in fact, I feel having a `pretrained_` prefix for everything that's already under a `pretrained` module is a bit redundant. @ZhitingHu Do you think we can remove the `pretrained_` prefix for everything except `pretrained_base`, e.g.:
> 
> * `pretrained_bert` to `bert`
> * `pretrained_xlnet_test` to `xlnet_test`

Yes, it is just the name change. I just want to make `texar-tf/pretrained` and `texar-pytorch/pretrained` be consistent with each other. If we want to remove `pretrained_` prefix, maybe we can change `pretrained_bert_tokenizer.py` to `bert_tokenzier.py` as well (?)

huzecong:
> maybe we can change pretrained_bert_tokenizer.py to bert_tokenzier.py as well

I'm in favor of this (granted that we will actually make such changes), especially since we call the tokenizers themselves `BERTTokenizer` and such, without the `Pretrained` prefix.

ZhitingHu:
Yes it's nice to keep the file name the same w/ the class name (i.e., no `pretrained_`).

Issue #41, opened 2019-06-14
Redesign module constructor interfaces


huzecong:
When subclassing existing modules, a common use case I've encountered is to automatically fill certain arguments of the super class constructor based on HParams values. By design, HParams are realized in the base class (`ModuleBase`), where the default values are filled in and user-provided values are type-checked. However, in order to call the base class constructor, all it arguments must be filled, and these arguments may rely on default HParams values, which are not available yet.

As a result, one must explicitly realize the HParams in the derived class. But there is no way to prevent the base class from realizing the HParams again, and in cases with multiple levels of inheritance, HParams realization could happen multiple times.

For example, let's say I'm writing a `CharCNN` module and want to subclass `texar.modules.encoders.Conv1DEncoder`. The signature of the constructor for `Conv1DEncoder` is: `__init__(self, in_channels: int, in_features: Optional[int] = None, hparams=None)`. I have a field named `"char_embed_dim"` in `default_hparams`, which I am using to fill the `in_channels` argument. To do that, I have to realize HParams.

One possible solution is to change the constructor of HParams: `__init__(self, hparams, default_hparams)`. If `hparams` is already an HParams instance, then we don't do anything. This way, when a subclass realizes HParams, it can pass the realized HParams into the super class constructor, thus prevent repeated realization.

gpengzhi:
See Zecong's comments in #208

PR #208, opened 2019-09-13
Refactor XLNet interface

gpengzhi:
Requested by Zhiting's comments in [PR in texar-tf](https://github.com/asyml/texar/pull/206)

Make the interface of `XLNetEncoder` be consistent with other modules.

### DIFF
--- a/texar/torch/modules/decoders/xlnet_decoder.py
+++ b/texar/torch/modules/decoders/xlnet_decoder.py
@@ -15,7 +15,6 @@
 from typing import Any, Dict, List, NamedTuple, Optional, Tuple, Type, Union
 
 import torch
-from torch import nn
 from torch.nn import functional as F
 
 from texar.torch.modules.decoders.decoder_base import DecoderBase
@@ -66,26 +65,12 @@ class XLNetDecoder(XLNetEncoder, DecoderBase[Optional[State], Output]):
             :meth:`default_hparams` for the hyperparameter structure
             and default values.
     """
-
+    _IS_DECODE = True
     # Variables persistent during decoding.
     _state_cache_len: int
     _state_recompute_memory: bool
     # required for recomputing memory
     _state_previous_inputs: List[torch.Tensor]
-
-    def __init__(self,
-                 pretrained_model_name: Optional[str] = None,
-                 cache_dir: Optional[str] = None,
-                 hparams=None):
-
-        super().__init__(pretrained_model_name=pretrained_model_name,
-                         cache_dir=cache_dir,
-                         hparams=hparams,
-                         init=False)
-
-        self.lm_bias = nn.Parameter(torch.zeros(self._hparams.vocab_size))
-
-        self.init_pretrained_weights()
 
     @staticmethod
     def default_hparams() -> Dict[str, Any]:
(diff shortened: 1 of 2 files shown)


gpengzhi:
> Looks good. I'm thinking, would this problem be solved if we have a mechanism that can:
> 
> * Add methods that is called prior to `__init__` (of the uppermost subclass).
> * Add methods that is called after `__init__` (of the uppermost subclass).
> 
> I feel that such mechanism could solve both this problem and the repeated-hparams initialization problem we've discussed before.
> 
> A hacky way of doing this is via metaclasses. When an instance of a class is created, the metaclass `__call__` method is invoked, which internally calls the class `__new__` and class `__init__` methods. We can create a custom metaclass for `ModuleBase` that constructs the `hparams` object before `__init__`, and allows sub-metaclasses to append method calls after `__init__`.
> 
> A diagram of instance creation is as follows, taken from this [very good article that explains it all](https://blog.ionelmc.ro/2015/02/09/understanding-python-metaclasses/):
> `

Thanks a lot. I think this modification is related to #41 I will think about it.

ZhitingHu:
Why is this PR related to "repeated-hparams initialization"?  In particular, what's the `init` argument  in the XLNetEncoder `__init__` for in the first place?

huzecong:
It's because these two problems have similar root causes.

**The `init` argument.** Our pre-trained mixin interfaces requires the concrete classes to call `init_pretrained_weights()` at the end of their `__init__` methods. This method is used to initialize the weights of the registered module parameters from the checkpoint file. 

`XLNetEncoder`, which inherits `PretrainedXLNetMixin`, follows this pattern. However, `XLNetDecoder` inherits `XLNetEncoder` (which is not a good design; but it works at least for now), but has an extra parameter `lm_bias` that has to be registered before calling `init_pretrained_weights()`, which happens with the `super().__init__` call. Due to PyTorch implementations, we can't register parameters before `super().__init__` (which internally calls `nn.Module.__init__`), so we had this `init` argument to control whether we call `init_pretrained_weights()` in `XLNetEncoder.__init__`.

**Relation to repeated-`hparams` creation.** The `hparams` issue was that: `hparams` is created in `ModuleBase.__init__`, which will be called in `super().__init__` of our built-in modules. However, certain base classes require additional arguments in their constructors (e.g., `DataBase` require a `data_source` argument), and some arguments cannot be constructed without knowing `hparams` values. As a results, the `hparams` object is constructed multiple times.

The `init` issue above could benefit from being able to add method calls (a call to `init_pretrained_weights()`) at the end of object initialization. The `hparams` issue could similarly benefit from being able to add calls before initialization (a call to construct `hparams`).

ZhitingHu:
Thanks for explaining, and the possible solution.

The solution can be a candidate we can apply at some point. Since it's a bit intricate, for now (when the issue has not widely occurred), let's just work around case by case.

One of our interface design "principles" could to some extent (not completely) avoid the issue: put hyperparameters in `hparams` instead of constructor arguments **whenever possible**. In this way, it's in general unlikely (though not impossible) that an hyperparam is in the `hparams` of a class but at the same time is an argument of `super().__init__`

It looks the `CharCNN` example in the issue https://github.com/asyml/texar-pytorch/issues/41 is out of the scope of the above principle though. It's reasonable that `in_channels` is `Conv1DEncoder.__init__` argument and `char_embed_dim` is in `CharCNN` `hparams`.

huzecong:
Yes, we could leave this for the future.

While the issue in the `CharCNN` example could be avoided with this principle, it only applies to values that fits in `hparams` -- and a data source is probably not one of them. There could also be logic that has to be done before calling the super class constructor.

PR #214, opened 2019-09-22
Update CI

gpengzhi:


huzecong:
Could you explain the reason behind this change?

gpengzhi:
My bad. Did not notice the change to .[extras]

PR #220, opened 2019-09-30
Resolve #218: Support soft ids as inputs for BERT/GPT2/RoBERTa/XLNet

### LINKED ISSUE #218: Support soft_ids as the input for BERTEncoder/GPT2Encoder/RoBERTaEncoder...
Currently, the `input` for these modules must be hard_ids (`[batch_size, max_time]`). We could support soft_ids (`[batch_size, max_time, vocab_size]`) as well so that the embedding is computed by `self.word_embedder(soft_ids=soft_ids)`.

gpengzhi:
Support `soft_ids` as inputs for `encoders`, `classifiers`, and `regressors`.

### DIFF
--- a/texar/torch/losses/adv_losses.py
+++ b/texar/torch/losses/adv_losses.py
@@ -35,6 +35,11 @@ def binary_adversarial_losses(
     r"""Computes adversarial losses of real/fake binary discrimination game.
 
     Example:
+
+    .. code-block:: python
+
+        # Using BERTClassifier as the discriminator, which can accept
+        # "soft" token ids for gradient backpropagation
         discriminator = tx.modules.BERTClassifier('bert-base-uncased')
 
         G_loss, D_loss = tx.losses.binary_adversarial_losses(
--- a/texar/torch/losses/adv_losses.py
+++ b/texar/torch/losses/adv_losses.py
@@ -27,17 +27,20 @@ __all__ = [
 ]
 
 
-__all__ = [
-    'binary_adversarial_losses',
-]
-
-
 def binary_adversarial_losses(
         real_data: torch.Tensor,
         fake_data: torch.Tensor,
         discriminator_fn: Callable[[torch.Tensor], MaybeTuple[torch.Tensor]],
         mode: str = "max_real") -> Tuple[torch.Tensor, torch.Tensor]:
     r"""Computes adversarial losses of real/fake binary discrimination game.
+
+    Example:
+        discriminator = tx.modules.BERTClassifier('bert-base-uncased')
+
+        G_loss, D_loss = tx.losses.binary_adversarial_losses(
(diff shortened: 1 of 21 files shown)


gpengzhi:
Resolve #218

gpengzhi:
> LGTM. What worries me is that our implementation of modules based on pre-trained stuff is a bit too repetitive, so that a seemingly small change would require modifying a bunch of files. This is also true for a lot of tests (not limited to pre-trained ones). Let's keep this in mind so we can improve this in the future.

This is a very good suggestion. For the code redundancy in the models, I feel that we can extract some task-specific headers for all or most of the pre-trained models. For the testing issue, some of the tests can be shared across different models. I think we can write a common test script that can be used by a group of classes for unit testing.

huzecong:
That sounds good. Let's open a separate issue to keep track of this, so we can merge this PR.

gpengzhi:
I will open an issue to keep track of this. @ZhitingHu suggested me to have two arguments (`token_ids` and `soft_token_ids`) instead of one (`inputs`) for these classes. I will change the interface a little bit accordingly.

texar/torch/losses/adv_losses.py
```diff
@@ -27,18 +27,21 @@
...
 def binary_adversarial_losses(
         real_data: torch.Tensor,
         fake_data: torch.Tensor,
         discriminator_fn: Callable[[torch.Tensor], MaybeTuple[torch.Tensor]],
         mode: str = "max_real") -> Tuple[torch.Tensor, torch.Tensor]:
     r"""Computes adversarial losses of real/fake binary discrimination game.
 
+    Example:
+        discriminator = tx.modules.BERTClassifier('bert-base-uncased')
+
```
ZhitingHu: Add comment: 

```
# Using BERTClassifier as the discriminator, which can accept "soft" token ids for gradient backpropagation 
```

texar/torch/modules/classifiers/gpt2_classifier.py
```diff
@@ -202,7 +202,10 @@ def forward(self,  # type: ignore
 
         Args:
             inputs: A 2D Tensor of shape `[batch_size, max_time]`,
-                containing the token ids of tokens in input sequences.
+                containing the token ids of tokens in input sequences or
+                a 3D Tensor of shape `[batch_size, max_time, vocab_size]`,
+                containing the weights (probabilities) used to mix the
+                embedding vectors.
```
ZhitingHu: Polished a bit:

```
Either a **2D Tensor** of shape `[batch_size, max_time]`, containing the ids of tokens in input sequences, or a **3D Tensor** of shape `[batch_size, max_time, vocab_size]` containing soft token ids (i.e., weights or probabilities) used to mix the embedding vectors.
```

huzecong:
Feel free to merge when you're done with the changes.

PR #241, opened 2019-11-01
Bugfix in RoBERTa

gpengzhi:
"10.xxxxx" => "10" and "xxxxx"


huzecong:
Why was it not caught by tests before? How did you find this?

gpengzhi:
> Why was it not caught by tests before? How did you find this?

No error was triggered before because `layer_name is not in attention_tensor_map`. Maybe it should raise a `NameError` if the variable with this name is not found. I found it when I added `SpanBERT`.

PR #249, opened 2019-11-19
Support PyTorch 1.3

### LINKED ISSUE #231: Support PyTorch 1.3.0 & Add it to CI

### LINKED ISSUE #234: A (potential bug) in data module
Run `seq2seq_attn` example.

```python
python prepare_data.py --data toy_copy

python seq2seq_attn.py --config-model config_model --config-data config_toy_copy
```
Have the following error. I believe there was no such error before.

```
step=0, loss=36.4250
step=50, loss=22.1697
step=100, loss=13.6045
step=150, loss=6.7436
step=200, loss=4.6603
step=250, loss=2.5102
step=300, loss=0.2983
Traceback (most recent call last):
  File "seq2seq_attn.py", line 180, in <module>
    main()
  File "seq2seq_attn.py", line 166, in main
    _train_epoch()
  File "seq2seq_attn.py", line 131, in _train_epoch
    for batch in data_iterator:
  File "/Users/pengzhi.gao/Desktop/my_git/texar-pytorch/texar/torch/data/data/data_iterators.py", line 426, in __next__
    not self.dataset.hparams.allow_smaller_final_batch):
AttributeError: '_SPDataLoaderIter' object has no attribute 'dataset'
```

### LINKED ISSUE #247: Bugs in data_iterators.py
https://github.com/asyml/texar-pytorch/blob/267bfd7f5fbb6c27e1a35c85aedab6edb124903b/texar/torch/data/data/data_iterators.py#L426

The code above should be:
···
not self._dataset.hparams.allow_smaller_final_batch):
···
from dataset to _dataset

gpengzhi:
resolve #247 
resolve #234 
resolve #231

### DIFF
--- a/texar/torch/data/data/data_iterators.py
+++ b/texar/torch/data/data/data_iterators.py
@@ -18,17 +18,18 @@ Various data iterator classes.
 # pylint: disable=protected-access
 
 from typing import (
-    Any, Callable, Dict, Generic, Iterable, Iterator, List, Optional,
-    Sequence, Tuple, TypeVar, Union, Mapping)
+    Dict, Iterable, Iterator, List, Optional, Sequence, Union, Mapping)
 
 import pkg_resources
 import torch
 from torch import __version__ as _torch_version  # type: ignore
 from torch.utils.data import DataLoader
-from torch.utils.data import sampler as torch_sampler
 
 from texar.torch.data.data.data_base import DatasetBase
 from texar.torch.data.data.dataset_utils import Batch
+from texar.torch.data.data.sampler import (
+    SamplerBase, SequentialSampler, RandomSampler, BufferShuffleSampler,
+    BatchingStrategy, DynamicBatchSampler)
 from texar.torch.utils.types import MaybeSeq
 from texar.torch.utils.utils import ceildiv, map_structure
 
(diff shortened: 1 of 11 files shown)


texar/torch/data/data/data_iterators.py
```diff
@@ -421,10 +433,16 @@ def __next__(self):
             batch = super().__next__()
             # Drop smaller final batch according to settings. Note that
             # `_batch_size` could be None if dynamic batching is used.
-            if (self._batch_size is not None and
-                    batch.batch_size < self._batch_size and
-                    not self.dataset.hparams.allow_smaller_final_batch):
-                raise StopIteration
+            if _torch_version >= pkg_resources.parse_version("1.3.0"):
```
huzecong: This introduces a large amount of code repetition, plus calling `pkg_resources.parse_version` and checking the version every time is wasteful. I think we can do better. For instance, create a wrapper class over the iterators for PyTorch 1.3 that adds the attribute names with underscore as properties.

It might also be a good idea to move certain classes to a separate file, since this file is getting a bit too long. We might also want to restructure the `data` folder.

huzecong: What I mentioned above is just one way of doing things, and I'm sure there are better ways. Please try to think of a clean and elegant way to resolve this problem.

gpengzhi:
Tested on `seq2seq_attn`, `gpt-2`, and `sentence_classifier` examples.

AvinashBukkittu:
Thanks for doing this! With new releases, Pytorch is making changes to a few attribute names in its core module, hence we end up making some updates to our data modules. LGTM.

PR #252, opened 2019-11-21
Fix bug in rnn

Codle:
When using rnn without sequence_length, the new tensor created for sequence_length don't have same device as inputs. It will cause the torch.utils.mask_sequences arguments have different device.


huzecong:
Thanks for catching this, and thanks for creating a pull request! I think there are also other tensor creation ops that needs fixing, for instance the one on line 282. Could you also fix those?

Codle:
OK, I will check other creation ops tomorrow.

huzecong:
Please also make sure that the code passes our integration tests, which includes formatting checks. Now it fails because you have a trailing whitespace.

Issue #246, opened 2019-11-17
Bugs in roberta RoBERTaTokenizer.encode_text


li3cmz:
https://github.com/asyml/texar-pytorch/blob/267bfd7f5fbb6c27e1a35c85aedab6edb124903b/texar/torch/data/tokenizers/roberta_tokenizer.py#L198-L210

When I use RoBERTaTokenizer.encode_text，
https://github.com/asyml/texar-pytorch/blob/267bfd7f5fbb6c27e1a35c85aedab6edb124903b/texar/torch/data/tokenizers/roberta_tokenizer.py#L92
The return id of cls_token or "spe_token" is None

PR #255, opened 2019-11-27
Update RoBERTa vocabulary files

gpengzhi:
```python
import torch
roberta = torch.hub.load('pytorch/fairseq', 'roberta.base')
roberta.eval()

tokens = roberta.encode('Hello world!')
print(tokens)  # [    0, 31414,   232,   328,     2]
```

```python
import texar.torch as tx
tokenizer = tx.data.RoBERTaTokenizer(pretrained_model_name='roberta-base')

input_ids, _ = tokenizer.encode_text('Hello world!', max_seq_length=5)
print(input_ids)  # [0, 31414, 232, 328, 2]
```

### DIFF
--- a/texar/torch/data/tokenizers/gpt2_tokenizer.py
+++ b/texar/torch/data/tokenizers/gpt2_tokenizer.py
@@ -70,6 +70,24 @@ class GPT2Tokenizer(TokenizerBase, PretrainedGPT2Mixin):
         'vocab_file': 'encoder.json',
         'merges_file': 'vocab.bpe',
     }
+    _VOCAB_FILE_MAP = {
+        'vocab_file': {
+            'gpt2-small': 'encoder.json',
+            'gpt2-medium': 'encoder.json',
+            'gpt2-large': 'encoder.json',
+            'gpt2-xl': 'encoder.json',
+            '117M': 'encoder.json',
+            '345M': 'encoder.json',
+        },
+        'merges_file': {
+            'gpt2-small': 'vocab.bpe',
+            'gpt2-medium': 'vocab.bpe',
+            'gpt2-large': 'vocab.bpe',
+            'gpt2-xl': 'vocab.bpe',
+            '117M': 'vocab.bpe',
+            '345M': 'vocab.bpe',
+        },
+    }
 
     def __init__(self,
                  pretrained_model_name: Optional[str] = None,
@@ -84,10 +102,13 @@ class GPT2Tokenizer(TokenizerBase, PretrainedGPT2Mixin):
         }
 
         if self.pretrained_model_dir is not None:
+            assert self.pretrained_model_name is not None
             vocab_file = os.path.join(self.pretrained_model_dir,
-                                      self._VOCAB_FILE_NAMES['vocab_file'])
+                                      self._VOCAB_FILE_MAP['vocab_file']
+                                      [self.pretrained_model_name])
             merges_file = os.path.join(self.pretrained_model_dir,
-                                       self._VOCAB_FILE_NAMES['merges_file'])
+                                       self._VOCAB_FILE_MAP['merges_file']
+                                       [self.pretrained_model_name])
             assert pretrained_model_name is not None
             if self._MAX_INPUT_SIZE.get(pretrained_model_name):
                 self.max_len = self._MAX_INPUT_SIZE[pretrained_model_name]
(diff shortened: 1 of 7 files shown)


gpengzhi:
resolve #246

texar/torch/data/tokenizers/gpt2_tokenizer.py
```diff
@@ -70,6 +70,24 @@ class GPT2Tokenizer(TokenizerBase, PretrainedGPT2Mixin):
...
+    _VOCAB_FILE_MAP = {
+        'vocab_file': {
+            'gpt2-small': 'encoder.json',
+            'gpt2-medium': 'encoder.json',
+            'gpt2-large': 'encoder.json',
+            'gpt2-xl': 'encoder.json',
+            '117M': 'encoder.json',
+            '345M': 'encoder.json',
+        },
+        'merges_file': {
```
ZhitingHu: what does `merges` mean here?

texar/torch/data/tokenizers/roberta_tokenizer.py
```diff
@@ -24,8 +24,7 @@
     'RoBERTaTokenizer',
 ]
 
-_GPT2_PATH = "https://dl.fbaipublicfiles.com/fairseq/gpt2_bpe/"
-_CHECKPOINT_FILES = ["encoder.json", "vocab.bpe"]
+_ROBERTA_PATH = "https://s3.amazonaws.com/models.huggingface.co/bert/"
```
ZhitingHu: Can we host the checkpoints by ourselves, on the asyml google drive?

texar/torch/data/tokenizers/gpt2_tokenizer.py
gpengzhi: BPE generates subword tokens by merging the most frequent consecutive byte pairs. This file contains all the merged byte pairs.

texar/torch/data/tokenizers/roberta_tokenizer.py
gpengzhi: We can. Do we currently have asyml google drive?

gpengzhi:
> `input_ids, _ = tokenizer.encode_text('Hello world!', max_seq_length=5)`
> 
> is `max_seq_length` necessary here? what's the result in this case without setting `max_seq_length`?

It will be

gpengzhi:
> `input_ids, _ = tokenizer.encode_text('Hello world!', max_seq_length=5)`
> 
> is `max_seq_length` necessary here? what's the result in this case without setting `max_seq_length`?

`max_seq_length` is used here just to show the same result between ours and theirs. If `max_seq_length` is not specified, it will be `[0, 31414, 232, 328, 2, 0, 0, 0, ..., 0]`(zero-pad up to the maximum sequence length).

texar/torch/data/tokenizers/roberta_tokenizer.py
ZhitingHu: > > `input_ids, _ = tokenizer.encode_text('Hello world!', max_seq_length=5)`
> > is `max_seq_length` necessary here? what's the result in this case without setting `max_seq_length`?
> 
> `max_seq_length` is used here just to show the same result between ours and theirs. If `max_seq_length` is not specified, it will be `[0, 31414, 232, 328, 2, 0, 0, 0, ..., 0]`(zero-pad up to the maximum sequence length).

If the user doesn't want padding, it's difficult for them to know the correct seq length just as in this case ('Hello world!' needs seq_length 5).. Can we add an argument (or allow a special value of `max_seq_length`) so that the user can get the encoded text without padding (and without need to specify `max-seq-length` explicitly)?

ZhitingHu:
> > `input_ids, _ = tokenizer.encode_text('Hello world!', max_seq_length=5)`
> > is `max_seq_length` necessary here? what's the result in this case without setting `max_seq_length`?
> 
> `max_seq_length` is used here just to show the same result between ours and theirs. If `max_seq_length` is not specified, it will be `[0, 31414, 232, 328, 2, 0, 0, 0, ..., 0]`(zero-pad up to the maximum sequence length).

If the user doesn't want padding, it's difficult for them to know the correct seq length just as in this case ('Hello world!' needs seq_length 5).. Can we add an argument (or allow a special value of max_seq_length) so that the user can get the encoded text without padding (and without need to specify max-seq-length explicitly)?

gpengzhi:
> > > `input_ids, _ = tokenizer.encode_text('Hello world!', max_seq_length=5)`
> > > is `max_seq_length` necessary here? what's the result in this case without setting `max_seq_length`?
> > 
> > 
> > `max_seq_length` is used here just to show the same result between ours and theirs. If `max_seq_length` is not specified, it will be `[0, 31414, 232, 328, 2, 0, 0, 0, ..., 0]`(zero-pad up to the maximum sequence length).
> 
> If the user doesn't want padding, it's difficult for them to know the correct seq length just as in this case ('Hello world!' needs seq_length 5).. Can we add an argument (or allow a special value of max_seq_length) so that the user can get the encoded text without padding (and without need to specify max-seq-length explicitly)?

There are two returns, `input_ids` and `input_mask`, for the function `encode_text`. If the user can compute the correct sequence length by aceessing `input_mask`. In this case, `input_mask` is `[1, 1, 1, 1, 1, 0, 0, ..., 0]`.

ZhitingHu:
pls try to pass the test and merge asap

gpengzhi:
I tried to disable `codecov/patch` because some of our tests (pre-trained tests) are done locally. `codecov/patch` is also reported to be not very reliable. Some other projects have met such issues (https://github.com/argoproj/argo-cd/pull/1926).  I think `codecov/project` is good enough to check the code quality for now. I tried many ways to disable `codecov/patch` but failed. It seems that someone also met a similar problem (https://community.codecov.io/t/cannot-disable-codecov-patch-check/682/4). Do you think it is reasonable to merge PRs regardless of the status of `codecov/patch`? @ZhitingHu

ZhitingHu:
As long as the build is "passing" we're good for now

gpengzhi:
https://github.com/pytorch/fairseq/issues/1432

PR #273, opened 2019-12-18
XLNet tokenizer bug

atif93:
The assertion should pass even if the `pretrained_model_name` is not passed.

Instead of checking if `pretrained_model_name` is not None, we have to check if `self.pretrained_model_name` is not None. This is because `pretrained_model_name` is an optional argument and `self.pretrained_model_name` is set accordingly if `pretrained_model_name` is None.


gpengzhi:
Thanks for this fix! I think lines 99 and 100 should also be fixed.

PR #274, opened 2019-12-18
Making the download_checkpoint pass even if the downloaded file isn't extracted into a new directory

atif93:
The download_checkpoint function assumes that the file will be extracted into a new directory but that isn't always the case.


gpengzhi:
Thanks. LGTM.

atif93:
Not sure what this codecov failure means. Can we still merge?

gpengzhi:
> Not sure what this codecov failure means. Can we still merge?

It is weird that `codecov/patch` appears again ... Let me investigate this at first.

PR #275, opened 2019-12-18
Fixing the Google Drive download bug

atif93:
When downloading from Google Drive we use the `file_id` for saving the download file name. But when deleting that same file after extracting in `download_checkpoint` method, we get the filename from the URL which can be different in many cases.

Also, in `_extract_google_drive_file_id` method, we should be able to extract the `file_id` even if there's no trailing `/` in the URL.

### DIFF
--- a/texar/torch/data/data_utils.py
+++ b/texar/torch/data/data_utils.py
@@ -33,6 +33,7 @@ __all__ = [
     "read_words",
     "make_vocab",
     "count_file_lines",
+    "get_filename"
 ]
 
 Py3 = sys.version_info[0] == 3
--- a/texar/torch/data/data_utils.py
+++ b/texar/torch/data/data_utils.py
@@ -311,6 +311,8 @@ def count_file_lines(filenames: MaybeList[str]) -> int:
 
 
 def get_filename(url: str) -> str:
+    r"""Extracts the filename of the downloaded checkpoint file from the URL.
+    """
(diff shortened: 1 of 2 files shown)


texar/torch/data/data_utils.py
```diff
@@ -305,3 +308,12 @@ def _count_lines(fn):
         filenames = [filenames]
     num_lines = np.sum([_count_lines(fn) for fn in filenames]).item()
     return num_lines
+
+
+def get_filename(url: str) -> str:
```
gpengzhi: Add `get_filename` to `__all__`.

texar/torch/modules/pretrained/pretrained_base.py
```diff
@@ -23,6 +23,7 @@
 from torch import nn
 
 from texar.torch.data.data_utils import maybe_download
+from texar.torch.data.data_utils import get_filename
```
gpengzhi: `from texar.torch.data.data_utils import get_filename, maybe_download`

texar/torch/data/data_utils.py
atif93: Added

texar/torch/modules/pretrained/pretrained_base.py
atif93: Done

PR #277, opened 2019-12-20
Adding BERT for MS-MARCO Passage re-ranking

### LINKED ISSUE #254: Support Re-ranking BERT
Passage Re-ranking with BERT

https://github.com/nyu-dl/dl4marco-bert

atif93:
Adding BERT fine-tuned on MS-MARCO for passage re-ranking task (https://arxiv.org/abs/1901.04085)

Since this is a pretrained classifier, we had to add final linear layer parameters in `PretrainedBERTMixin`. Based on the `pretrained_model_name`, the weights of the final classifier layer will be loaded if they are present.

resolve #254

### DIFF
--- a/texar/torch/data/tokenizers/bert_tokenizer.py
+++ b/texar/torch/data/tokenizers/bert_tokenizer.py
@@ -74,6 +74,10 @@ class BERTTokenizer(PretrainedBERTMixin, TokenizerBase):
         'scibert-scivocab-cased': 512,
         'scibert-basevocab-uncased': 512,
         'scibert-basevocab-cased': 512,
+
+        # BERT for MS-MARCO
+        'bert-msmarco-base': 512,
+        'bert-msmarco-large': 512,
     }
     _VOCAB_FILE_NAMES = {'vocab_file': 'vocab.txt'}
     _VOCAB_FILE_MAP = {
@@ -98,6 +102,10 @@ class BERTTokenizer(PretrainedBERTMixin, TokenizerBase):
             'scibert-scivocab-cased': 'vocab.txt',
             'scibert-basevocab-uncased': 'vocab.txt',
             'scibert-basevocab-cased': 'vocab.txt',
+
+            # BERT for MS-MARCO
+            'bert-msmarco-base': 'vocab.txt',
+            'bert-msmarco-large': 'vocab.txt',
         }
     }
 
--- a/texar/torch/data/tokenizers/bert_tokenizer.py
+++ b/texar/torch/data/tokenizers/bert_tokenizer.py
@@ -76,8 +76,8 @@ class BERTTokenizer(PretrainedBERTMixin, TokenizerBase):
         'scibert-basevocab-cased': 512,
 
         # BERT for MS-MARCO
-        'bert-msmarco-base': 512,
-        'bert-msmarco-large': 512,
+        'bert-msmarco-nogueira19-base': 512,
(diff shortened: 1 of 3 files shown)


atif93:
Wanted to get an idea of what you guys think of this design for loading a pretrained BERTClassifier config.  
Next I'll be writing an example to test that we can reproduce the results of the paper.

texar/torch/modules/classifiers/bert_classifier.py
```diff
@@ -71,6 +71,8 @@ def __init__(self,
 
         super().__init__(hparams=hparams)
 
+        self.load_pretrained_config(pretrained_model_name, cache_dir)
```
gpengzhi: If I understand it correctly, will `load_pretrained_config` and `init_pretrained_weights` be called twice (once in `BERTClassifier`, and once in `BERTEncoder`)?

gpengzhi: If that is the case, we probably should not load the pre-trained weights in `self.encoder (BERTEncoder)`.

texar/torch/data/tokenizers/bert_tokenizer.py
```diff
@@ -74,6 +74,10 @@ class BERTTokenizer(PretrainedBERTMixin, TokenizerBase):
         'scibert-scivocab-cased': 512,
         'scibert-basevocab-uncased': 512,
         'scibert-basevocab-cased': 512,
+
+        # BERT for MS-MARCO
+        'bert-msmarco-base': 512,
+        'bert-msmarco-large': 512,
```
hunterhector: This won't be the last/best Bert model for MS-Marco, so probably we'd come up with more specific names, say bert-msmarco-nguyen2019

texar/torch/modules/pretrained/bert.py
```diff
@@ -97,6 +98,16 @@ class PretrainedBERTMixin(PretrainedMixin, ABC):
         * ``scibert-basevocab-cased``: Cased version of the model trained on
           the original BERT vocabulary.
 
+    * **BERT for MS-MARCO**: proposed in (`Nogueira et al`. 2019)
+      `Passage Re-ranking with BERT`_. A BERT model fine-tuned on MS-MARCO
+      (Nguyen et al., 2016) dataset. It's the best performing model (on Jan 8th
```
gpengzhi: `Nguyen et al`

texar/torch/modules/classifiers/bert_classifier.py
atif93: Discussed offline.
We can pass `pretrained_model_name` as None while instantiating the encoder in `BERTClassifier`.

texar/torch/data/tokenizers/bert_tokenizer.py
atif93: Sure let me change that

texar/torch/modules/classifiers/bert_classifier.py
gpengzhi: If you set `pretrained_model_name` and `pretrained_model_name` in `hparams` to be `None`, `BERTEncoder` won't load the pre-trained weights.

atif93: Made both the changes.

texar/torch/data/tokenizers/bert_tokenizer.py
atif93: Changed the names

PR #292, opened 2020-02-05
Travis update

gpengzhi:
1. Fix the existing issues in our CI:

```
both matrix and jobs given, matrix overwrites jobs
```

2. Update the versions of `mypy`, `pylint`, and `flake8` & fix those errors accordingly.


examples/bert/data/download_glue_data.py
```diff
@@ -129,19 +129,21 @@ def main(arguments) -> None:
...
+            if not os.path.exists("data/MRPC"):
+                subprocess.run("mkdir data/MRPC", shell=True, check=False)
             subprocess.run(
                 'wget -P data/MRPC/ https://dl.fbaipublicfiles.com/senteval/senteval_data/msr_paraphrase_train.txt',
-                shell=True)
+                shell=True, check=False)
             subprocess.run(
                 'wget -P data/MRPC/ https://dl.fbaipublicfiles.com/senteval/senteval_data/msr_paraphrase_test.txt',
-                shell=True)
+                shell=True, check=False)
```
huzecong: Do we really want `check=False`? We'll be needing these files, so a non-zero return code would mean that the file is not downloaded, which would cause problems. To me it's better if such error is not ignored.

texar/torch/modules/classifiers/conv_classifiers.py
```diff
@@ -163,7 +163,8 @@ def default_hparams() -> Dict[str, Any]:
 
     def forward(self,  # type:ignore
                 input: torch.Tensor,
-                sequence_length: Union[torch.LongTensor, List[int]] = None,
+                sequence_length: Optional[Union[torch.LongTensor,
+                                                List[int]]] = None,
```
huzecong: We don't allow `sequence_length` to be `List[int]` in some methods (such as the one above, in `losses/rewards.py` `discount_reward`). Let's keep the behavior consistent. If we decide that `sequence_length` can be `List[int]` everywhere, we might as well add a type alias for it and add some utility functions to transform it into `LongTensor`.

huzecong:
I think this PR is good to go. We may address other problems in a separate one.

PR #299, opened 2020-02-24
Add ELMo modules 

gpengzhi:
Add texar-styled ELMo encoder adapted from allennlp. The corresponding tokenizer will be in another PR.

Resolve some comments in #298 

I checked the implementation of `ELMo` in `allennlp`, It seems that they used customized LSTM such that we cannot use our `LSTM` module to implement it directly. And the `Highway` module they used is different from our `HighwayWrapper`. I feel that it is better to directly use their implementations, and the correctness of the implementation is guaranteed by their unit tests. Please let me know your thought @huzecong

### DIFF
--- a/docs/code/modules.rst
+++ b/docs/code/modules.rst
@@ -64,6 +64,11 @@ Encoders
 .. autoclass:: texar.torch.modules.BERTEncoder
     :members:
 
+:hidden:`ELMoEncoder`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. autoclass:: texar.torch.modules.ELMoEncoder
+    :members:
+
 :hidden:`RoBERTaEncoder`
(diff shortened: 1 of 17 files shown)


huzecong:
I agree that their implementations are a bit different and it's more reliable to use their code as is, but I don't think I would approve of merging into master at this point.

True, if I were to choose between installing `allennlp` along with a plethora of dependencies, and installing this version of `texar`, I would choose `texar`. However, `ELMo` is not the only module worth using in our package. To introduce 3k lines of code and an additional dependency is a bit too much for just one functionality. I still believe it is possible to somehow rewrite the module to use more existing parts in our package, but I agree that it would mean a lot of work and might not be worth it at this point.

Speaking of too much code, although not related to this PR, I think it's better to move the `*_test.py` files out of the `texar` folder, and move them to a separate `tests/` folder. This way the end user doesn't have to install the test files when they do `pip install texar-pytorch`. @ZhitingHu What do you think?

PR #300, opened 2020-02-25
Add SpanBERT module

### LINKED ISSUE #230: Support SpanBERT

gpengzhi:
resolve #230 

SpanBERT extends BERT by (1) masking contiguous random spans, rather than random tokens, and (2) training the span boundary representations to predict the entire content of the masked span, without relying on the individual token representations within it. Differing from the standard BERT, the SpanBERT model does not use segmentation embedding.

resolve comments in #240


texar/torch/modules/encoders/bert_encoder.py
```diff
@@ -328,7 +331,9 @@ def forward(self,  # type: ignore
                                      dtype=torch.int64)
         pos_embeds = self.position_embedder(sequence_length=pos_length)
 
-        if self.segment_embedder is not None:
+        if self.segment_embedder is not None and \
+                self.pretrained_model_name is not None and \
+                not self.pretrained_model_name.startswith('spanbert'):
```
huzecong: Why do we perform this check here? Wouldn't it make more sense to not construct a `segment_embedder` for SpanBERT?

gpengzhi: I agree.

PR #302, opened 2020-03-03
Add sequence_tagging example

gpengzhi:
Adapted from [sequence_tagging](https://github.com/asyml/texar/tree/master/examples/sequence_tagging) in `texar-tf`.


examples/sequence_tagging/README.md
```diff
@@ -0,0 +1,29 @@
...
+The model will begin training, and will evaluate on the validation data periodically, and evaluate on the test data after the training is done. 
+
+## Results ##
+
+The results on validation and test data is:
+
+|       |   precision   |  recall  |    F1    |
+|-------|----------|----------|----------|
+| valid |  91.98   |  93.30   |  92.63   |
+| test  |  87.39   |  89.78   |  88.57   |
```
huzecong: Are these the previous results, or results from the PyTorch version?

examples/sequence_tagging/conll_reader.py
```diff
@@ -0,0 +1,266 @@
...
+
+import re
+
+from collections import defaultdict
+
+import numpy as np
+
+import texar.torch as tx
+
+# pylint: disable=redefined-outer-name, unused-variable
```
huzecong: Why are these necessary? I can imagine `unused-variable` being required for the constants, but why `redefined-outer-name`? It's also better practice to add a corresponding `pylint: enable` comment after where it's no longer required.

examples/sequence_tagging/conll_reader.py
```diff
@@ -0,0 +1,266 @@
...
+NUM_CHAR_PAD = 2
+
+UNK_WORD, UNK_CHAR, UNK_NER = 0, 0, 0
+PAD_WORD, PAD_CHAR, PAD_NER = 1, 1, 1
+
+# Regular expressions used to normalize digits.
+DIGIT_RE = re.compile(r"\d")
+
+
+def create_vocabs(train_path, dev_path, test_path, normalize_digits=True,
```
huzecong: Please add type annotations for functions.

examples/sequence_tagging/ner.py
```diff
@@ -0,0 +1,256 @@
...
+
+scale = np.sqrt(3.0 / EMBEDD_DIM)
+word_vecs = np.random.uniform(
+    -scale, scale, [len(word_vocab), EMBEDD_DIM]).astype(np.float32)
+if config.load_glove:
+    word_vecs = construct_init_word_vecs(word_vocab, word_vecs, glove_dict)
+
+scale = np.sqrt(3.0 / CHAR_DIM)
+char_vecs = np.random.uniform(
+    -scale, scale, [len(char_vocab), CHAR_DIM]).astype(np.float32)
```
huzecong: Consider moving these into `main` as well.

examples/sequence_tagging/ner.py
```diff
@@ -0,0 +1,256 @@
...
+config: Any = importlib.import_module(args.config)
+
+device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+
+train_path = os.path.join(args.data_path, args.train)
+dev_path = os.path.join(args.data_path, args.dev)
+test_path = os.path.join(args.data_path, args.test)
+embedding_path = os.path.join(args.data_path, args.embedding)
+
+EMBEDD_DIM = config.embed_dim
```
huzecong: Why `EMBEDD` instead of `EMBED`?

examples/sequence_tagging/conll_reader.py
```diff
@@ -0,0 +1,266 @@
...
+            embedding = glove_dict[word.lower()]
+        else:
+            embedding = None
+
+        if embedding is not None:
+            word_vecs[index] = embedding
+    return word_vecs
+
+
+class CoNLLReader:
```
huzecong: Can we modify this to use `tx.data.Dataset`? This would eliminate the need for separate `read_data` and `iterate_batch` methods.

examples/sequence_tagging/conll_reader.py
```diff
@@ -0,0 +1,266 @@
...
+            word_ids.append(self.__word_vocab[word])
+
+            ner_tags.append(ner)
+            ner_ids.append(self.__ner_vocab[ner])
+
+        return NERInstance(Sentence(words, word_ids, char_seqs, char_id_seqs),
+                           ner_tags, ner_ids)
+
+
+class NERInstance:
```
huzecong: This and `Sentence` below can be changed to `NamedTuple`s, which also supports custom methods.

examples/sequence_tagging/conll_reader.py
```diff
@@ -0,0 +1,266 @@
...
+                cid_inputs[i, c, len(cids):] = PAD_CHAR
+            cid_inputs[i, inst_size:, :] = PAD_CHAR
+            nid_inputs[i, :inst_size] = nids
+            nid_inputs[i, inst_size:] = PAD_NER
+            masks[i, :inst_size] = 1.0
+
+        yield wid_inputs, cid_inputs, nid_inputs, masks, lengths
+
+
+def load_glove(filename, emb_dim, normalize_digits=True):
```
huzecong: I thought we had a `load_glove` method inside `tx.data`? What are the differences here?

examples/sequence_tagging/ner.py
```diff
@@ -0,0 +1,256 @@
...
+            hparams={"rnn_cell_fw": config.cell, "rnn_cell_bw": config.cell})
+
+        self.dropout_1 = nn.Dropout(p=0.33)
+        self.dense_1 = nn.Linear(in_features=2 * config.hidden_size,
+                                 out_features=config.tag_space)
+        self.dropout_2 = nn.Dropout(p=(1 - config.keep_prob))
+        self.dense_2 = nn.Linear(in_features=config.tag_space,
+                                 out_features=len(ner_vocab))
+
+    def forward(self, inputs, chars, targets, masks, seq_lengths, mode):
```
huzecong: Type annotations.

examples/sequence_tagging/conll_writer.py
```diff
@@ -0,0 +1,41 @@
...
+    def __init__(self, i2w, i2n):
+        self.__source_file = None
+        self.__i2w = i2w
+        self.__i2n = i2n
+
+    def start(self, file_path):
+        self.__source_file = open(file_path, 'w', encoding='utf-8')
+
+    def close(self):
+        self.__source_file.close()
```
huzecong: We could change these to `__enter__` and `__exit__` and use the `with writer.open(path) as f` context manager pattern. It's also fine to keep it as is if you're more comfortable with this.

huzecong (changes requested):

examples/sequence_tagging/README.md
gpengzhi: from the PyTorch version

PR #311, opened 2020-04-23
Set train mode after evaluation in training

### LINKED ISSUE #310: Switch between model.train() and model.eval()
https://github.com/asyml/texar-pytorch/blob/95f3d644680656f38b7b1779f2ba184ef9e1f81a/examples/bert/bert_classifier_main.py#L164

Hello, 
     I find that after doing "_eval_epoch()"，you don't switch model.eval() to model.train(). Wouldn't that be a problem?

AvinashBukkittu:
This PR 
- Fixes a bug in BERT and GPT2 examples of not setting mode to train after evaluation

Fixes #310

### DIFF
--- a/examples/bert/README.md
+++ b/examples/bert/README.md
@@ -120,14 +120,24 @@ After convergence, the evaluation performance is around the following. Due to ce
 initialization of the classification layer), the evaluation accuracy is reasonable as long as it's `>0.84`.
 An example output is as follows:
 ```
-Using cached pre-trained BERT checkpoint from texar_download/BERT/bert-base-uncased.
-INFO:root:step: 50; loss: 0.448667
-INFO:root:step: 100; loss: 0.400019
-INFO:root:step: 150; loss: 0.340314
-INFO:root:step: 200; loss: 0.151271
-INFO:root:step: 250; loss: 0.093740
-INFO:root:step: 300; loss: 0.161118
-INFO:root:eval accu: 0.8554; loss: 0.4524; nsamples: 408
+Using cached pre-trained BERT checkpoint from /home/centos/texar_data/BERT/bert-base-uncased.
+INFO:root:step: 50; loss: 0.646327
+INFO:root:step: 100; loss: 0.281063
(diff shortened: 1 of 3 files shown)


huzecong:
Have you reran the fine-tuning experiments? It would be best to ensure that we're doing it right, and that the results improve compared to when we did it wrong. We might also want to update the README with the new outputs.

AvinashBukkittu:
> Have you reran the fine-tuning experiments? It would be best to ensure that we're doing it right, and that the results improve compared to when we did it wrong. We might also want to update the README with the new outputs.

I did not completely re-run the experiments. I re-ran until the first `_eval_epoch` call in training, checked the value of mode and current dataset in the iterator right after this call (the mode was `eval` and current dataset was eval dataset, which is wrong). 

Sure, I can update the Readme with the new results.

huzecong:
Hmm, interesting... The loss values now are much (one order of magnitude) lower than before starting from iter 250, and it seems a bit weird to me. Also, I checked the code at the commit where the README was modified to include the results:
https://github.com/asyml/texar-pytorch/blob/e4b68188388dbaa07791528d066b410a6f838de7/examples/bert/config_data.py#L9

It seems that `eval_step` was set to -1, which means no evaluation is performed until the end of training. In this case, the previous results were not incorrect (although the implementation was still faulty).

AvinashBukkittu:
> Hmm, interesting... The loss values now are much (one order of magnitude) lower than before starting from iter 250, and it seems a bit weird to me. Also, I checked the code at the commit where the README was modified to include the results:
> https://github.com/asyml/texar-pytorch/blob/e4b68188388dbaa07791528d066b410a6f838de7/examples/bert/config_data.py#L9
> 
> It seems that `eval_step` was set to -1, which means no evaluation is performed until the end of training. In this case, the previous results were not incorrect (although the implementation was still faulty).

We start noticing a stark difference for 300th iteration. Agree, previously, after first call `_eval_epoch()`, the mode was still set to `eval`

Issue #322, opened 2020-09-15
A logging bug in executor after close_files


hunterhector:
## The Problem
Executor will call [`close_files`](https://github.com/asyml/texar-pytorch/blob/master/texar/torch/run/executor.py#L1766)  in some cases, and it will see the `self._log_destination` list to be empty. 

If we call [`_open_files`](https://github.com/asyml/texar-pytorch/blob/master/texar/torch/run/executor.py#L1733) afterward, which assumes the `self._log_destination` to be of the same length as `self.log_destination`. This will raise an `IndexError`. 

## Possible Fix
Either change the `close_file` logic to make sure the list won't be zero or change the `_open_files` logic to make sure the we recreate the list to avoid `IndexError`

PR #323, opened 2020-09-15
Fixing index error after close files and clear `_log_destination`

hunterhector:
This PR fixes https://github.com/asyml/texar-pytorch/issues/322

## The fix
At the time of `_open_files`, we check whether the length of the destination is zero, and fill in the List if needed.

## The test
In the executor tests, we called `test` after `train` in all the tests. In this way, we ensure the `_files_opened` to have both `True` and `False` value, to allow us to test both branches in the `_open_files` function.

### DIFF
--- a/tests/run/executor_test.py
+++ b/tests/run/executor_test.py
@@ -19,6 +19,7 @@ import tempfile
 import unittest
 from pathlib import Path
 import os
+import sys
 from typing import List, Dict, Tuple
 
 import torch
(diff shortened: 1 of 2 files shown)


huzecong:
For some reason adding `executor.test()` doesn't raise exceptions when I test locally without applying the fix... Strange.

This fix definitely works, however I think the easier thing to do is to simply not set `_log_destinations` to `[]`. Since we maintain `_files_opened`, we don't need to check whether `_log_destinations` has length 0.

hunterhector:
> For some reason adding `executor.test()` doesn't raise exceptions when I test locally without applying the fix... Strange.

You are right indeed, to make the bug appear we need to add a file log in the log_destination parameter. I have added that in the newer commit.

> This fix definitely works, however I think the easier thing to do is to simply not set `_log_destinations` to `[]`. Since we maintain `_files_opened`, we don't need to check whether `_log_destinations` has length 0.

Added in the newer commit, see if that works.

huzecong (approved):

PR #332, opened 2021-02-16
Add an extra function tokenize_with_span to WordpieceTokenizer

jieralice13:


gpengzhi:
Could you add the corresponding unit test for the added function (here)[https://github.com/asyml/texar-pytorch/blob/master/tests/data/tokenizers/bert_tokenizer_utils_test.py]? BTW, the new function looks almost the same as the existing `tokenize` function. Is it possible to refactor the existing function to achieve your expected functionality?

gpengzhi:
Thanks for the modifications! Could you also fix the CI error?

jieralice13:
> Thanks for the modifications! Could you also fix the CI error?

Thanks, I've fixed the CI error.

tests/data/tokenizers/bert_tokenizer_utils_test.py
```diff
@@ -62,13 +63,24 @@ def test_wordpiece_tokenizer(self):
...
 
         self.assertListEqual(tokenizer.tokenize(""), [])
+        self.assertListEqual(tokenizer.tokenize("", with_span=True), [])
 
         self.assertListEqual(
             tokenizer.tokenize("unwanted running"),
             ["un", "##want", "##ed", "runn", "##ing"])
+        self.assertListEqual(
+            tokenizer.tokenize("unwanted running", with_span=True), [
+                ('un\t0\t2'), ('##want\t2\t6'),
```
gpengzhi, line 73: Such string structure looks a little weird to me. But if this is the expected structure what @hunterhector or @ZhitingHu wanted for the project, I am OK to merge this PR.

hunterhector: It looks like this implementation tries to get around the typing issue by converting integers to string. I agree with pengzhi that this looks weird. The behavior of this function would be quite hard to understand for the users.

The typing issues should be directly fixable without doing this.
1. This change [here](https://github.com/asyml/texar-pytorch/pull/332/commits/4c290a890f5c9697485bf231af05f33ec8ef3ac2#diff-e5889098da537d273c5122fc8b281b9738038774b79baeada05c7e4141c73760) forgot to import Union
2. You should be able to make the return type of `TokenizerBase` more general, the implemented function can return a more specific type if they wanted to. (https://mypy.readthedocs.io/en/latest/common_issues.html#incompatible-overrides). I.e. you can return `Union[List[str], Tuple[List[str, int, int]]]` in the base, and the implementer class can return either the same, or simply `List[str]`.  
3. You cannot extend the return type from `List[Union[str, Tuple[str, int, int]]]` to `List[Tuple[str, int, int]]`, nor can you use `List[Tuple[str, int, int]]` as a return type.  Because, in Python, mutable List is invariant, which means when you extend it, you cannot change the inner type of it.  The element type inside the list must be the same.  Now you are changing the inner type from `Union[str, Tuple[str, int, int]]` to `Tuple[str, int, int]`, making it more specific. This is only allowed for `covariant` types. I think here you can change the parent type to `Union[List[str], List[Tuple[str, int, int]]`, and then extend it with `Union[List[Tuple[str, int, int]]`, because `Union` is clearly covariant (i.e. you can change the types inside Union to be more specific) in return types.

Even with very complex typing cases, it is probably still better to use `type: ignore` sometimes than using string to represent numbewrs.

texar/torch/data/tokenizers/tokenizer_base.py
```diff
@@ -350,7 +350,8 @@ def split_on_tokens(tok_list, string):
         tokenized_text = split_on_tokens(added_tokens, text)
         return tokenized_text
 
-    def _map_text_to_token(self, text: str, **kwargs) -> List[str]:
+    def _map_text_to_token(self, text: str, **kwargs) -> \
+        Union[List[str], List[Tuple[str, int, int]]]:
```
hunterhector, line 354: Please update the docstring to reflect this too.

texar/torch/data/tokenizers/bert_tokenizer.py
```diff
@@ -166,7 +166,8 @@ def __init__(self,
         self.wordpiece_tokenizer = WordpieceTokenizer(vocab=self.vocab,
                                                       unk_token=self.unk_token)
 
-    def _map_text_to_token(self, text: str) -> List[str]:  # type: ignore
+    def _map_text_to_token(self, text: str) -> \
+        Union[List[str], List[Tuple[str, int, int]]]:  # type: ignore
```
hunterhector, line 170: this `type: ignore` need to be on the previous line.

texar/torch/data/tokenizers/bert_tokenizer.py
```diff
@@ -166,7 +166,8 @@ def __init__(self,
...
         split_tokens = []
```
hunterhector, line 171: To avoid the typing issue here, you can create two different lists inside the `if` and `else` block:
`split_tokens: List[str] = []` and `split_token_span: List[Tuple[str, int, int]] = []`, and return then separately

hunterhector:
There is a migration process from travis org to travis com, for the PRs at this moment, please check the build status manually https://travis-ci.org/github/asyml/texar-pytorch/pull_requests

texar/torch/data/tokenizers/bert_tokenizer.py
```diff
@@ -166,7 +166,8 @@ def __init__(self,
         self.wordpiece_tokenizer = WordpieceTokenizer(vocab=self.vocab,
                                                       unk_token=self.unk_token)
 
-    def _map_text_to_token(self, text: str) -> List[str]:  # type: ignore
+    def _map_text_to_token(self, text: str) -> \
```
hunterhector, line 169: add # type: ignore at line 169

texar/torch/data/tokenizers/bert_tokenizer.py
jieralice13: I think _map_text_to_token should only return split_tokens as List[str]. It doesn't need span info. My thought is that it is a within class function and called by "map_text_to_token", then "map_token_to_id" calls "map_text_to_token". These functions don't need span info.

jieralice13: Perhaps separate it out as another function "tokenize_with_span" is better.

gpengzhi:
@jieralice13 Instead of waiting for the response from CI, you can actually run all CI tests (mypy, unittest, ...) locally. You could check the correctness of your implementation before pushing codes to GitHub. This might save you time when you contribute to the repo in the future.

texar/torch/data/tokenizers/tokenizer_base.py
hunterhector, line 354: Can you reflect the change in the return type in the docstring?

texar/torch/data/tokenizers/bert_tokenizer_utils.py
```diff
@@ -223,34 +223,34 @@ def __init__(self, vocab: Dict[str, int],
...
+        r"""Tokenizes a piece of text into its word pieces with span info.
 
         This uses a greedy longest-match-first algorithm to perform tokenization
         using the given vocabulary.
 
         For example:
             input = "unaffable"
-            output = ["un", "##aff", "##able"]
+            output = [("un", 0, 2), ("##aff", 2, 5), ("##able", 5, 9)]
+                  when with_span is True
```
hunterhector, line 235: You don't have `with_span` as the parameter anymore.

texar/torch/data/tokenizers/bert_tokenizer_utils.py
```diff
@@ -265,13 +265,24 @@ def tokenize(self, text: str) -> List[str]:
...
+                output_tokens_and_span.append((self.unk_token, 0, len(chars)))
             else:
-                output_tokens.extend(sub_tokens)
+                output_tokens_and_span.extend(sub_tokens_and_span)
+
+        return output_tokens_and_span
+
+    def tokenize(self, text: str) -> List[str]:
+        """
+        Tokenizes a piece of text into its word pieces, remove span info.
```
hunterhector, line 280: The term "remove span info" is confusing for users who don't know the internal calls. You should simply describe what this does. You can point to the other function if you would like to tell the users how to preserve the spans.

You should also copy the `longest-match-first algorithm` part here.

hunterhector (commented):
Documentation changes required.

hunterhector (approved):
LGTM

PR #334, opened 2021-04-06
To get the __len__, should pass self._current_dataset_name to _valida…

limberc:
Pass self._current_dataset_name to self._validate_dataset_name. Or it would get error when calling __len__.

### DIFF
--- a/texar/torch/data/data/data_iterators.py
+++ b/texar/torch/data/data/data_iterators.py
@@ -94,7 +94,6 @@ if _torch_version >= pkg_resources.parse_version("1.2.0"):  # PyTorch 1.2.0 +
     from texar.torch.data.data.data_iterators_utils import \
         TexarMultiProcessingDataLoaderIter as _MultiProcessingDataLoaderIter
 
-
     class _DataLoaderIter(_BaseDataLoaderIter):
         r"""Iterates once over the DataLoader's dataset. This is almost
         identical to PyTorch
--- a/texar/torch/data/data/data_iterators.py
+++ b/texar/torch/data/data/data_iterators.py
@@ -43,7 +43,6 @@ __all__ = [
 # `Dict` is invariant, `Mapping` is not.
 DatasetsType = Union[Mapping[str, DatasetBase], MaybeSeq[DatasetBase]]
 
-
 # pylint: disable=ungrouped-imports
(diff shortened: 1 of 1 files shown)


gpengzhi:
Thank you very much for contributing to `texar-pytorch`! Could you fix the CI error at first? It seems to be a pylint error.
'''
************* Module texar.torch.data.data.data_iterators
texar/torch/data/data/data_iterators.py:559:0: C0301: Line too long (91/80) (line-too-long))
'''

texar/torch/data/data/data_iterators.py
```diff
@@ -95,6 +94,7 @@ def _move_fn(x):
     from texar.torch.data.data.data_iterators_utils import \
         TexarMultiProcessingDataLoaderIter as _MultiProcessingDataLoaderIter
 
+
```
huzecong, line 97: These whitespace changes are probably why the CI lint failed. Let's get rid of these and only have changes relevant to the issue in this PR.

huzecong (approved):
Thanks for the contribution! I think this is a good fix. Please fix the CI issues before merging!

Related but definitely out of the scope for this PR: down the road we should probably refactor the data interfaces a bit, from hindsight the current implementation is not very intuitive.

texar/torch/data/data/data_iterators.py
limberc: Wow, thanks for the advice.

limberc:
I have deleted the whitespace according to @huzecong

texar/torch/data/data/data_iterators.py
```diff
@@ -549,7 +555,7 @@ def __iter__(self) -> Iterator[Batch]:
         return self.get_iterator()
 
     def __len__(self):
-        return len(self._datasets[self._validate_dataset_name(None)])
+        return len(self._datasets[self._validate_dataset_name(self._current_dataset_name)])
```
huzecong, line 558: Sorry but I don't think you've solved all of the lint errors. Can you only keep the change on this line?

Also, I think this line is too long. Our line limit is 80 characters. Could you wrap this?

huzecong (changes requested):

huzecong:
This was addressed in #351, so I'm closing this PR. Thanks anyway for your contribution!

PR #336, opened 2021-04-15
remove key

hunterhector:
This PR removes a private key for security reasons.


huzecong:
You probably want to rewrite history using `bfg` and force-push. Otherwise the key stays in commit history forever.

hunterhector:
> You probably want to rewrite history using `bfg` and force-push. Otherwise the key stays in commit history forever.

Good suggestion, I think the key is actually just a public key so there is no real security thread. But we are not using it anymore so we remove it. Do you think it is necessary in danger of making problems on git history?

huzecong:
Ah okay then it's probably fine. 😛

Issue #343, opened 2021-07-22
Wordpiece tokenizer tokenize_with_span(text) should return the span begin and end relative to the text 


jennyzhang-petuum:
Wordpiece tokenizer `tokenize_with_span(text)` method should return the span begin and end relative to the text rather than the token. Since `whitespace_tokenize` is called on input `text` within the function to get the tokens, we should return span begin and end relative to the text directly to avoid another round of tokenization.

gpengzhi:
Resolved

PR #344, opened 2021-07-22
fix wordpiece tokenizer to return relative position

jennyzhang-petuum:
The PR solves #343 
Test for bert_tokenizer_utils.py has been updated to the new span.


hunterhector (approved):
LGTM

PR #351, opened 2021-12-26
Fix `DataIterator.__len__`

### LINKED ISSUE #350: Issue in DataIterator.__len__
The current implementation of [`DataIterator.__len__`](https://github.com/asyml/texar-pytorch/blob/5d67ae957763a3683ce8dd5e4d1208cc9fdb4d33/texar/torch/data/data/data_iterators.py#L551) is
```python
def __len__(self):
        return len(self._datasets[self._validate_dataset_name(None)])
```
This only works for `DataIterator` that only contains a single dataset. Otherwise it will trigger a [ValueError](https://github.com/asyml/texar-pytorch/blob/5d67ae957763a3683ce8dd5e4d1208cc9fdb4d33/texar/torch/data/data/data_iterators.py#L511).
For example, if I run the following code snippet:
```python
dataiterator = DataIterator(datasets={'ds1': dataset1, 'ds2': dataset2})
dataiterator.switch_to_dataset('ds2')
print(len(dataiterator))
```
The expected behavior would be to print out the length of `dataset2`, but instead it raised a [ValueError](https://github.com/asyml/texar-pytorch/blob/5d67ae957763a3683ce8dd5e4d1208cc9fdb4d33/texar/torch/data/data/data_iterators.py#L511).
Maybe we should update the implementation of `DataIterator.__len__` to something like:
```python
def __len__(self):
        return len(self._datasets[self._validate_dataset_name(self._current_dataset_name)])
```
in order to avoid this problem.

mylibrar:
This PR fixes #350. There is also a previous PR(#334) that tried to address the same problem.
- Update [`DataIterator.__len__`](https://github.com/asyml/texar-pytorch/blob/5d67ae957763a3683ce8dd5e4d1208cc9fdb4d33/texar/torch/data/data/data_iterators.py#L551)
- Add `pip install six` to `setup.py` and `requirements.txt`. Otherwise it will trigger an [error](https://github.com/mylibrar/texar-pytorch/runs/4637728147?check_suite_focus=true#step:8:146).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,7 +52,7 @@ jobs:
           pip install --progress-bar off .[extras]
           if [[ ${{ matrix.torch-version }} == "1.4.0" ]]; then pip install pylint==2.4.4 flake8==3.7.9; fi
           if [[ ${{ matrix.torch-version }} != "1.0.1" ]]; then pip install mypy==0.790; fi
-          pip install pytest six
+          pip install pytest
           pip install --progress-bar off coverage codecov
       - name: Linting pylint and flake8 if torch version is 1.4.0
         run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,7 +52,7 @@ jobs:
           pip install --progress-bar off .[extras]
           if [[ ${{ matrix.torch-version }} == "1.4.0" ]]; then pip install pylint==2.4.4 flake8==3.7.9; fi
           if [[ ${{ matrix.torch-version }} != "1.0.1" ]]; then pip install mypy==0.790; fi
-          pip install pytest
+          pip install pytest six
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -52,7 +52,7 @@ jobs:
           pip install --progress-bar off .[extras]
           if [[ ${{ matrix.torch-version }} == "1.4.0" ]]; then pip install pylint==2.4.4 flake8==3.7.9; fi
           if [[ ${{ matrix.torch-version }} != "1.0.1" ]]; then pip install mypy==0.790; fi
-          pip install pytest
+          pip install pytest six
```
huzecong, line 55: I think you should list `six` as a requirement in `requirements.txt` and in `install_requires` in `setup.py` instead of hard-coding it here.

huzecong (approved):

PR #352, opened 2022-02-15
Workflow and code update for numpy versions from 1.15 to 1.21

### LINKED ISSUE #333: Compatibility check with numpy>=1.20
Numpy has made a big release of version [1.20.0](https://numpy.org/doc/stable/release/1.20.0-notes.html). Texar is not fully tested with the new release. One particular problem is the newly introduced type annotations for Numpy functions. This change will at least make mypy raise type errors.

To update to the new numpy version, the mypy-raised errors should be reviewed and fixed accordingly.

To get the list of typing errors, run `mypy .` with `numpy >=1.20`, or check the previously workflow log: https://github.com/asyml/texar-pytorch/runs/2197557226?check_suite_focus=true

### LINKED ISSUE #341: Compatibility update with pytorch>=1.7
Texar is not fully tested with the new release of PyTorch after 1.7.0. One particular problem is the newly introduced type annotations for PyTorch. This change will at least make mypy raise type errors.

To update to the new PyTorch version, the mypy-raised errors should be reviewed and fixed accordingly.

To get the list of typing errors, run mypy . with PyTorch >=1.7.1, or check the previously workflow log: https://github.com/mylibrar/texar-pytorch/runs/3001591121?check_suite_focus=true

tanyuqian:
This PR fixes https://github.com/asyml/texar-pytorch/issues/333 and fixes https://github.com/asyml/texar-pytorch/issues/341

Code is updated to pass the mypy test with numpy>=1.20 (which fixes #333)

Workflow is updated to enumerate numpy versions (from 1.15 to 1.21), and further include pytorch version (1.7.1 and 1.8.1), which fixes #341 

Note: In [mypy.ini](https://github.com/tanyuqian/texar-pytorch/blob/master/mypy.ini), I changed ```warn_unused_ignores``` to ```False``` ([here](https://github.com/tanyuqian/texar-pytorch/blob/master/mypy.ini#L2)) for the enumeration of numpy versions. Not sure if there are better ways to do it like in [mypy-torch](https://github.com/tanyuqian/texar-pytorch/blob/master/mypy.ini#L13). Looks adding a same line under [[mypy-numpy]](https://github.com/tanyuqian/texar-pytorch/blob/master/mypy.ini#L20) doesn't work.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,7 +14,7 @@ jobs:
       matrix:
         python-version: [ 3.6, 3.7 ]
         torch-version: [ 1.0.1, 1.1.0, 1.2.0, 1.3.0, 1.4.0]
-        numpy-version: [ 1.21 ]
+        numpy-version: [ 1.19, 1.21 ]
         exclude:
           - python-version: 3.6
             numpy-version: 1.21
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,7 +14,7 @@ jobs:
       matrix:
         python-version: [ 3.6, 3.7 ]
         torch-version: [ 1.0.1, 1.1.0, 1.2.0, 1.3.0, 1.4.0]
-        numpy-version: [ 1.20.3, 1.21 ]
+        numpy-version: [ 1.19, 1.20.3, 1.21 ]
         exclude:
           - python-version: 3.6
             numpy-version: 1.21
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,12 +14,10 @@ jobs:
       matrix:
         python-version: [ 3.6, 3.7 ]
         torch-version: [ 1.0.1, 1.1.0, 1.2.0, 1.3.0, 1.4.0]
-        numpy-version: [ 1.19, 1.21 ]
+        numpy-version: [ 1.22 ]
         exclude:
           - python-version: 3.6
-            numpy-version: 1.20.3
-          - python-version: 3.6
-            numpy-version: 1.21
+            numpy-version: 1.22
           - python-version: 3.6
             torch-version: 1.0.1
           - python-version: 3.6
@@ -31,22 +29,13 @@ jobs:
         include:
           - python-version: 3.8
             torch-version: 1.7.1
-            numpy-version: 1.19
-          - python-version: 3.8
-            torch-version: 1.7.1
-            numpy-version: 1.21
+            numpy-version: 1.22
           - python-version: 3.8
             torch-version: 1.8.1
-            numpy-version: 1.19
-          - python-version: 3.8
-            torch-version: 1.8.1
-            numpy-version: 1.21
+            numpy-version: 1.22
           - python-version: 3.9
             torch-version: 1.8.1
-            numpy-version: 1.19
-          - python-version: 3.9
-            torch-version: 1.8.1
-            numpy-version: 1.21
+            numpy-version: 1.22
 
     steps:
       - uses: actions/checkout@v2
@@ -111,7 +100,7 @@ jobs:
           pip install --progress-bar off -r requirements.txt
           pip install --progress-bar off .[extras]
           pip install --progress-bar off -r docs/requirements.txt
-          pip install --progress-bar off --upgrade numpy==1.21
+          pip install --progress-bar off --upgrade numpy==1.22
       - name: Build Docs
         run: |
           cd docs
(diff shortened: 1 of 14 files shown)


hunterhector:
> Code is updated to pass the mypy test with numpy>=1.20
> 
> Workflow is updated to enumerate numpy versions (from 1.15 to 1.21)
> 
> Note: In [mypy.ini](https://github.com/tanyuqian/texar-pytorch/blob/master/mypy.ini), I changed `warn_unused_ignores` to `False` ([here](https://github.com/tanyuqian/texar-pytorch/blob/master/mypy.ini#L2)) for the enumeration of numpy versions. Not sure if there are better ways to do it like in [mypy-torch](https://github.com/tanyuqian/texar-pytorch/blob/master/mypy.ini#L13). Looks adding a same line under [[mypy-numpy]](https://github.com/tanyuqian/texar-pytorch/blob/master/mypy.ini#L20) doesn't work.

Don't know why the config does not work in the my-numpy section, does @huzecong have any ideas?

huzecong:
I think the `[mypy-torch]` section was for the PyTorch stubs files, not for code that calls PyTorch functions. I don't think there's any way to achieve the latter.

I'd imagine the `type: ignore` comments are added under a certain version of numpy, and does not apply to other versions? I think we can just run mypy/pylint on a single version.

hunterhector:
I see, so if I understand correctly, there are `type:ignore` comments in numpy and they use certain version of `numpy`/`mypy` combination that passes the tests. We will probably need to find the right version.

huzecong:
I don't think mypy checks numpy code since we've set `follow_imports = skip` in `mypy-numpy`. What I was referring to is Texar code that interacts with numpy. I believe we have some `type: ignore` comments on lines where we call numpy functions / interact with numpy arrays, and we can do type-checking on linting only on a single numpy version.

hunterhector:
OK, now I think I understand. And I think we are doing or planning to do mypy check with only certain versions. 

Now the fix would probably be, don't set `warn_unused_ignores` to False, fix a numpy version during mypy check, and remove unnecessary `type: ignore` when needed.

mypy.ini
```diff
@@ -1,5 +1,5 @@
 [mypy]
-warn_unused_ignores = True
+warn_unused_ignores = False
```
hunterhector, line 2: probably should revert this

.github/workflows/main.yml
```diff
@@ -54,14 +71,15 @@ jobs:
           if [[ ${{ matrix.torch-version }} != "1.0.1" ]]; then pip install mypy==0.790; fi
           pip install pytest
           pip install --progress-bar off coverage codecov
+          pip install --progress-bar off --upgrade numpy==${{ matrix.numpy-version }}
       - name: Linting pylint and flake8 if torch version is 1.4.0
         run: |
           if [[ ${{ matrix.torch-version }} == "1.4.0" ]]; then pylint texar/ examples/; fi
           if [[ ${{ matrix.torch-version }} == "1.4.0" ]]; then flake8 texar/ examples/; fi
-      - name: Linting with mypy if torch versoin is not 1.0.1
+      - name: Linting with mypy if torch versoin is not 1.0.1 and numpy>=1.20
```
hunterhector, line 79: maybe ==1.21`?

tanyuqian: now it tests numpy 1.20 and 1.21. do we want to test 1.21 only?

mypy.ini
tanyuqian: oh I forgot to push this change. updated.

.github/workflows/main.yml
hunterhector: I think type checking against one version should be good enough. In the current setting, newer version of numpy (say 1.3) may introduce different types which can break some tests)

tanyuqian: got it! updated.

hunterhector (approved):

hunterhector:
Thanks a lot for fixing this!

hunterhector (commented):
I realize two more things

1. Now numpy is on 1.22, would that be working?
2. We have only updated the test workflow versions, but not the actual install version check (i.e https://github.com/asyml/texar-pytorch/blob/master/requirements.txt and https://github.com/asyml/texar-pytorch/blob/master/setup.py#L35)

tanyuqian:
> I realize two more things
> 
> 1. Now NumPy is on 1.22, would that also work for us? For example, some of Forte check fail because of this: [asyml/forte#634](https://github.com/asyml/forte/issues/634)
> 2. We have only updated the test workflow versions, but not the actual install version check (i.e https://github.com/asyml/texar-pytorch/blob/master/requirements.txt and https://github.com/asyml/texar-pytorch/blob/master/setup.py#L35)

1. sure. numpy==1.22 for python 3.8 and 3.9 added. (numpy==1.22 requires python>=3.8)
2. install version check updated.

hunterhector:
thanks a lot

